PR #24, opened 2014-12-17
add PoC cmd converting HCL to JSON

josephholsten:
This adds an `hcl2json` command which reads HCL from STDIN and writes indented JSON to STDOUT.

At the moment it is not friendly at all, and probably is only useful for people who are testing HCL code changes. Before merging, it should be a friendly tool to allow people who :heart: HCL to use it in environments where importing this package is unwieldy.

Some things it should probably do:
- [x] `hcl2json`, `hcl2json -h` should print usage
- [x] `hcl2json foo.hcl` should read from `foo.hcl`
- [x] `hcl2json -` should read from STDIN
- [x] refactor to extract a CLI struct for testing
- [x] add tests for the above functionality
- build through gox?

Fixes https://github.com/hashicorp/hcl/issues/32

### DIFF
--- a/cmd/hcl2json/main.go
+++ b/cmd/hcl2json/main.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/hashicorp/hcl"
+)
+
+func main() {
+	d, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
+		log.Fatalf("err: %s", err)
+	}
+
+	var obj interface{}
+	err = hcl.Decode(&obj, string(d))
+	if err != nil {
+		log.Fatalf("err: %s", err)
+	}
+
+	out, err := json.MarshalIndent(obj, "", "  ")
+	if err != nil {
+		log.Fatalf("err: %s", err)
+	}
+
+	fmt.Println(string(out))
+}
--- a/cmd/hcl2json/main.go
+++ b/cmd/hcl2json/main.go
@@ -11,21 +11,36 @@ import (
 )
 
 func main() {
-	args := os.Args
+	cli := &CLI{
+		Args:   os.Args,
+		Stdin:  os.Stdin,
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	}
+	os.Exit(cli.Run())
+}
 
-	if 2 != len(args) {
-		usage()
+type CLI struct {
+	Args   []string
+	Stdin  *os.File
+	Stdout *os.File
+	Stderr *os.File
+}
+
+func (c *CLI) Run() int {
+	if 2 != len(c.Args) {
+		c.Usage()
 	}
 
-	file := args[1]
+	file := c.Args[1]
 
 	var d []byte
 	var err error
 
 	if "-h" == file {
-		usage()
+		return c.Usage()
 	} else if "-" == file {
-		d, err = ioutil.ReadAll(os.Stdin)
+		d, err = ioutil.ReadAll(c.Stdin)
 	} else {
 		d, err = ioutil.ReadFile(file)
 	}
@@ -40,15 +55,17 @@ func main() {
 		log.Fatalf("err: %s", err)
 	}
 
-	out, err := json.MarshalIndent(obj, "", "  ")
+	res, err := json.MarshalIndent(obj, "", "  ")
 	if err != nil {
 		log.Fatalf("err: %s", err)
 	}
 
-	fmt.Println(string(out))
+	fmt.Fprintln(c.Stdout, string(res))
+
+	return 0
 }
 
-func usage() {
-	fmt.Fprintln(os.Stderr, "usage: hcl2json <file>")
-	os.Exit(1)
+func (c *CLI) Usage() int {
+	fmt.Fprintln(c.Stderr, "usage: hcl2json <file>")
+	return 1
 }
--- a/cmd/hcl2json/main.go
+++ b/cmd/hcl2json/main.go
@@ -11,7 +11,25 @@ import (
 )
 
 func main() {
-	d, err := ioutil.ReadAll(os.Stdin)
+	args := os.Args
+
+	if 2 != len(args) {
+		usage()
+	}
+
+	file := args[1]
+
+	var d []byte
+	var err error
+
+	if "-h" == file {
+		usage()
+	} else if "-" == file {
+		d, err = ioutil.ReadAll(os.Stdin)
+	} else {
+		d, err = ioutil.ReadFile(file)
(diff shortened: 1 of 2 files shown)


eropple:
Hey--

So I tried to use `hcl2json` to reverse-engineer the Terraform HCL format in order to work on [Terraframe](http://github.com/eropple/terraframe). The sample on the [file provisioner page](https://www.terraform.io/docs/provisioners/file.html) looks like this when fed through `hcl2json`:

```json
{
  "resource": [
    {
      "aws_instance": [
        {
          "web": [
            {
              "provisioner": [
                {
                  "file": [
                    {
                      "destination": "/etc/myapp.conf",
                      "source": "conf/myapp.conf"
                    }
                  ]
                },
                {
                  "file": [
                    {
                      "destination": "/etc",
                      "source": "conf/configs.d"
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

I have known-good TF input of similar structure that looks like this:

```json
{
  "resource": {
    "aws_instance": {
      "web": {
        "tags": {
          "Name": "web"
        }
      }
    }
  }
}
```


I still haven't figured out how `provisioner` is _supposed_ to look, but if you look at `resource` and `aws_instance`, you can see `hcl2json` appears to be adding a bunch of extra arrays to the tree where they're not supposed to be. I don't know if this is TF magic or a bug in `hcl2json`, but I was directed here by @mitchellh so I figured I'd make sure everything was cool. :)

josephholsten:
This is TF magic. Try running JSON through `hcl2json` to see how HCL parses into data. That output is what I'd expect to see.

eropple:
Gotcha, thanks.

virtuald:
+1, add some tests. This is a useful tool.

virtuald:
Interesting, it's worth noting that in at least one case, the output of hcl2json does not match the expected input from the fixtures. In particular, the terraform_heroku example does not match. This is probably similar to @eropple's concern.

As maintainer of pyhcl, I can attest that it has similar problems due to the ambiguities in hcl. It would be nice to get this clarified.

virtuald:
hcl2json's output of terraform_heroku.hcl:

```
{
  "config_vars": [
    {
      "FOO": "bar"
    }
  ],
  "name": "terraform-test-app"
}
```
pyhcl's hcltool output (which is the same as that checked into the repository):
```
{
    "config_vars": {
        "FOO": "bar"
    },
    "name": "terraform-test-app"
}
```

josephholsten:
@virtuald I've just refactored this to enable tests. Now I get to knock the rust off my go testing.

josephholsten:
all right, I've added some tests for the four use cases. This should be ready for code review and possibly merge.

solarce:
Just wanted to add my two ¢ here that it would be great to see this merged and also have the ambiguity around handling of multiple keys with the same name cleared up, e.g. https://github.com/virtuald/pyhcl/issues/1, as this is going to be the common case for any complex Terraform configuration files.

All it takes is two `aws_security_group` or `aws_instance` resource definitions in the same configuration file and you've got this "issue", as right now pyhcl's implementation will collapse them.

See https://gist.github.com/solarce/39d634fe5f6b3ae494c6 for a quick example

virtuald:
Yes, having this merged would be good -- but I would feel much better about this if the tests did tests against *all* of the existing fixtures -- just like I do with pyhcl.

virtuald:
Still waiting on something like this.

josephholsten:
@sethvargo what's remaining for this? Anything I can do to help?

@virtuald would you be satisfied with tests against everything in [`test-fixtures/`](https://github.com/hashicorp/hcl/tree/master/test-fixtures)? I'm not entirely sure why that's necessary as part of this tool, would it make more sense to just do these tests in [`decoder_test.go`](https://github.com/hashicorp/hcl/blob/master/decoder_test.go)?

virtuald:
@josephholsten Well, I'm not a maintainer of this repo, so technically my opinion doesn't count... 

As far as the fixtures go, it seems like it would make sense that for each of the fixtures, you compare the output of hcl2json to see if it actually matches the json file in the fixture directory as a sanity check.

jefferai:
There's another example of this tool adding arrays where they shouldn't be in https://github.com/hashicorp/vault/issues/582

jbaxleyiii:
I would also love to see this documented somewhere. I'm looking to write a node parser for HCL but am unsure of the mappings right now

kvz:
Same problem for me where `infra.tf` with

```hcl
variable "TSD_AWS_ACCESS_KEY" {
  description = "TSD_AWS_ACCESS_KEY"
}
```

Results in `infra.tf.json` with

```json5
{
  "variable": [
    {
      "TSD_AWS_ACCESS_KEY": [
        {
          "description": "TSD_AWS_ACCESS_KEY"
        }
      ]
    }
  ]
}
```

That Terraform trips over with a `root.Variable: not an object type for map (ValueTypeList)`

Then I tried with @virtuald's [hcltool](https://github.com/virtuald/pyhcl) which return JSON that Terraform said was valid:

```json5
{
    "variable": {
        "TSD_AWS_ACCESS_KEY": {
            "description": "TSD_AWS_ACCESS_KEY"
        }
    }
}
```

virtuald:
It would be great to get this finished off.

scalp42:
To be clear we can't convert the json outputted by to hcl correct?

The pain point for us is the templating of `*.tf` files (meaning ERBs for example) and I'm not exactly sure how other folks are doing it (besides storing vars in Consul backend for example).

josephholsten:
@scalp42 yes, at the moment this codebase is simply showing the internal data structures as JSON, not converting to the JSON that would generate those data structures.

snakescott:
Any ETA on hcl2json? I'm trying to get started with terraform, and want to both (a) write idiomatic config in hcl (.tf, not .tfjson) and (b) programmatically validate my config with python. hcl2json seems like the missing piece to get that workflow running.

sl1pm4t:
+1 to subscribe

kvz:
Hey there, you can use the HCL printer (and `terraform fmt`) to do this now. @acconut and I took the liberty of wrapping a few of HashiCorp's libraries and release `json2hcl` (and hcl2json via `-reverse`) binaries for this. In the offchance this is still useful to somebody: https://github.com/kvz/json2hcl

josephholsten:
@kvz woo! I'll give that a look soon. Closing in the assumption that we can move any issues over there.

kevinburkenotion:
Sorry - what is the "HCL printer" ?

Acconut:
HCL printer (see https://godoc.org/github.com/hashicorp/hcl/hcl/printer) is Go package which allows you to transform the in-memory representation of a HCL structure into the corresponding textual format. It's similar to Go's `json.Marshal` function but for HCL.

PR #36, opened 2015-08-31
Added hcl to json example for nested objects

timgent:
It was unclear how nested object example in hcl would map to json. Readme now updated to include an example of the json equivalent to a hcl nested object

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,16 @@ variable "ami" {
     description = "the AMI to use"
 }
 ```
+This would be equivalent to the following json:
+``` json
+{
+  "variable": {
+      "ami": {
+          "description": "the AMI to use"
+        }
+    }
+}
+```
 
 ## Thanks
 


jonashaag:
:+1:

sheerun:
Could you add example when multiple variables are defined?

sheerun:
e.g. 

```
resource "foo" "bar" {

}

resource "foo" "fiz" {

}
```

timgent:
I've rebased this to master so ready to merge but not sure how the hcl you've provided would map to json. Do you have the equivalent in json?

sheerun:
I have no idea

jen20:
Thanks for this @timgent! It certainly makes the usage of JSON a bit more obvious. I don't know what's up with Appveyor here, presumably the base was already broken. I'm not too worried though since the diff clearly only affects the README.

PR #37, opened 2015-09-05
json: support psuedo-comments

glenjamin:
As discussed in https://github.com/hashicorp/terraform/issues/784

Any key beginning with `//` will be ignored by the parser. I went with this format as it's one I've used in other projects, but there's plenty of room for bikeshedding on the actual logic for exclusion.

The approach I took was the have the JSON parser function do a recursive walk of the parsed `hcl.Object` and throw away any matching pairs. I think it could have used https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating to optimise a bit, but this seemed somewhat opaque.

I think it would also be possible to introduce another lexing token `COMMENT_STRING`, which behaves exactly as `STRING` in the value position, but causes the parser to skip in the key position - this would remove the need for an additional pass.

I'm pretty new to go, so please be as ruthless/picky as you like about style etc in any reviews.


glenjamin:
I also managed to get the alternative implementation working where the yacc parser is now comment-aware https://github.com/glenjamin/hcl/tree/json-comments-parser

While doing this I noticed that I seem to be running a newer version of `go tool yacc` than `hcl` was last generated with - and notably there is now an `ErrorVerbose` flag that can be set which includes the expected/unexpected token in syntax errors.

jen20:
Hi @glenjamin! Sorry this has taken a (very) long time to get to. The parser has been rewritten away from yacc very shortly after this was submitted, but I don't think anyone has has audited the older issues and pull requests for a while. I think we're pretty much set on maintaining compliance with the JSON standard in that form of HCL - if you need comments I'd recommend using the native HCL syntax which is designed for human rather than machine-centric use.

glenjamin:
@jen20 Thanks for the update - would you be open to specifying a top-level property that will be guaranteed to be ignored? so it can be used for header info like "this file was generated from XXX"

Issue #38, opened 2015-09-25
Decode into pointer fields in-place.


ianremmler:
It would be nice if decodePtr() would operate on the thing being pointed to rather than decoding into a new concrete value and assigning its address to the pointer.

I've found this behaviour handy using encoding/json, for example.  Say I have a struct prepopulated with defaults.  The struct corresponds to options nested under a key in a config file.  I can create an instance of an anonymous struct with a field whose name refers to that key and value is a pointer to my original struct.  I can then decode into the anonymous struct instance, and only fields present in the config file data will be updated, leaving the default fields intact.

PR #39, opened 2015-09-25
Decode into non-nil pointer's existing value.

ianremmler:
If decoding into a pointer, and the pointer points to a value, decode
into the exitsing value.  Otherwise, decode into a new value and point
to it.

Addresses issue #38.


langmartin:
https://github.com/hashicorp/hcl/pull/275 contains this change, slightly rebased to fit the shape of the code. Thanks for your fix, @ianremmler! fixes #38

PR #43, opened 2015-10-13
Support null values & bool values in arrays

yudai:
Hi, thank you for the convenient library.

This PR has two commits:

* Support null values
    * Currently HCL doesn't support `null` values. However, some configuration files need to support `null` values to JSON-compatible objects and arrays. This commit adds null value support to HCL.
* Allow bool values in arrays
    * This commit makes bool values to be able to exist in arrays.


jen20:
Hi @yudai! Thanks for submitting a pull request for this. I'd agree that we should probably support boolean values in arrays, and possibly also null values. Unfortunately this PR predates work which completely changed the underlying mechanism for lexing and parsing, meaning this won't apply.

If you'd like to update this to work with the current parser we can merge it - if you don't have time I'll try to pick it up at some point in future.

nethack42:
Hi guys, 

is there any update on supporting boolean values within lists?

PR #48, opened 2015-11-10
Decode heredocs in HCL

rgabo:
This PR adds back support for heredoc decoding by implementing the following changes:

 * Scan source starting with `<` (less-than sign) instead of treating it an illegal char.
 * Ensure that `<` is followed by another `<` to form heredoc: `<<MARKER`
 * Ensure that a marker is defined on the first line of the heredoc.
 * Scan everything until `MARKER` is detected on a newline and tokenize.
 * Decode heredoc tokens as `string`s and strip heredoc markers when decoding value.

I added an extra test for when there is no heredoc marker. Also made sure that `multiline.hcl` is the valid HCL and `multiline_bad.hcl` is the invalid HCL test fixture. The latter fixture has only one leading `<` to start the heredoc.

Hope this helps downstream people feel warm and fuzzy :smile:

Please comment and review according to how a Hashicorp PR should look like and I'll fix.


mitchellh:
Thanks! I just got back to a computer. I actually have a fulling working local unstaged copy :( So I'll commit that but I'll pull in your tests that you added on top of this to make sure things are extra tested.

I'm on the last part right now which is making the printer package work.

rgabo:
@mitchellh no worries, this was more of a fun exercise for me. There's actually an issue in the PR where the scanner misses the marker so I'll fix that first.

Feel free to close the PR itself and looking forward to your changeset :)

mitchellh:
See #49, I pulled in your tests. Thanks!

rgabo:
Awesome, thanks!

PR #51, opened 2015-11-12
parser: fix syntax error on missing comma

fatih:
Fixes hashicorp/terraform#3877

### DIFF
--- a/hcl/parser/parser.go
+++ b/hcl/parser/parser.go
@@ -248,16 +248,22 @@ func (p *Parser) listType() (*ast.ListType, error) {
 		Lbrack: p.tok.Pos,
 	}
 
+	needComma := false
 	for {
 		tok := p.scan()
 		switch tok.Type {
 		case token.NUMBER, token.FLOAT, token.STRING:
+			if needComma {
+				return nil, fmt.Errorf("unexpected token: %s. Expecting %s", tok.Type, token.COMMA)
+			}
+
 			node, err := p.literalType()
 			if err != nil {
 				return nil, err
 			}
 
 			l.Add(node)
+			needComma = true
 		case token.COMMA:
 			// get next list item or we are at the end
 			// do a look-ahead for line comment
@@ -271,6 +277,8 @@ func (p *Parser) listType() (*ast.ListType, error) {
 				}
 			}
 			p.unscan()
+
+			needComma = false
 			continue
 		case token.BOOL:
 			// TODO(arslan) should we support? not supported by HCL yet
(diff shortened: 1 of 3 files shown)


sethvargo:
@fatih so this fixes the issue :smile: but do we want to fix it?

sethvargo:
I kinda like not having to specify commas... /cc @mitchellh

fatih:
@sethvargo just added a test case for it. I think the new parser should be backwards compatible. This also fixes like cases for `["123" "456"]` which were not reported as errors. For multiline lists we might decide to neglect it, but as said I'm usually in favor of respecting backwards compatibility :)

sethvargo:
@fatih is it breaking backwards compat? Having commas still works and not having them also works :smile:

fatih:
@sethvargo oh :) If I didn't miss read here https://github.com/hashicorp/terraform/issues/3877 you're saying that it *errors* on` 0.6.6`, which I've assumed that `0.6.6` uses the old parser

sethvargo:
@fatih haha no problem. Let me clarify:

- The current version of Terraform 0.6.6, which uses the old parser, requires commas to separate values in arrays.
- The master branch of Terraform (0.6.7.dev), which uses the new parser, does not require commas to separate values in arrays.

```javascript
// Works in 0.6.6
inline = [
  "foo",
  "bar",
]

// Works in 0.6.7
inline = [
  "foo",
  "bar",
]
```

```javascript
// Does not work in 0.6.6
inline = [
  "foo"
  "bar"
]

// Works in 0.6.7
inline = [
  "foo"
  "bar"
]
```

Does that help clarify?

fatih:
Alright this now makes sense :+1:  So when I mean we should respect backwards compatibility, I was saying we should return an error for this case:

```
// Does not work in 0.6.6
inline = [
  "foo"
  "bar"
]
```

Right now, without this fix, the following also works (which imho should return also an error):

```
inline = ["foo" "bar"]
```

The parser doesn't check the newlines in any case. Having the comma is better for us, because it would make the parser more simpler and in the long term more stable. If we allow multiline lists without commas, we need to have fix it by also respecting newlines. This again, in my opinion, would complicate the parser.

I'm open to changes though :) I'm kinda following the philosophy of Go, basically having strict rules, explicitly over implicitly, and so on.

sethvargo:
Fair enough. I can deal :smile:

mitchellh:
I don't want to introduce new syntax into HCL and I prefer having commas. My preference is that we support trailing commas (which we do), but not the no comma case. This looks good to me. Thanks!

PR #52, opened 2015-11-12
Add travis.yml file

fatih:
We have many tests, having this would simplify contributing also easier
because Github would show us the CI result on PR's.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+sudo: false
+language: go
+go: 1.5


fatih:
@sethvargo we need to enable this on travis. I don't have any permission, but seeing you changing this file: https://github.com/hashicorp/envconsul/blob/master/.travis.yml I thought you could do it :)

mitchellh:
Thanks! I turned it on.

PR #53, opened 2015-11-13
ast: add Rewrite() to rewrite AST

fatih:
With the Walk function it's not easy to rewrite the node as we don't
have any kind of reference to the parent. If we want to rewrite a given
AST, we have to manually traverse it as Walk is not usable.

The new Rewrite function enables us to rewrite the AST easily by
returning the modified node for each traversed node. This idea was also
talked here:
https://groups.google.com/forum/#!topic/golang-nuts/cRZQV36IckM
extensively.

### DIFF
--- a/hcl/ast/ast_test.go
+++ b/hcl/ast/ast_test.go
@@ -2,6 +2,7 @@ package ast
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/hcl/hcl/token"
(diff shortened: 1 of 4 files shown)


mitchellh:
Can we just make `Walk` also rewrite? If it returns a non-nil node then it will replace that node? The logic is so similar I fear we'd have to duplicate it in so many places. Thoughts?

fatih:
I actually wanted to change `Walk`, but feared that it might break some other packages as it's public right now (even though it's new). 

Right now with `Walk` if you return a true it'll traverse till the end, when it's false it stops traversing. I can easily change the passed `fn` signature that it returns a node and not a bool and delete the current `Rewrite`. 

This would just require the user of this function to return always the node, instead of just signaling it with bool, like `true` or `false`. Somehow I feel this is not easy to use, but I don't know :) If you say it's ok I'll change the signature of `Walk` to:

```go
type WalkFunc func(Node) Node

func Walk(node Node, fn WalkFunc) Node {}
```

Example usage (without rewriting anything):

```go
Walk(node, func(n Node) Node {
	fmt.Printf("Position: %d\n", n.Pos())
	return n
})
```

Example usage (rewrite all keys to `foo`):

```go
// rewritten contains the new modified AST tree
rewritten = Walk(node, func(n Node) Node {
	switch t := n.(type) {
	case *ObjectKey:
		t.Token.Text = "foo"
		n = t
	}
	return n
})

```

mitchellh:
Working with ASTs in the past, being able to halt walking is actually really important, so I think the signature would have to change to `(Node, bool)`, where the semantics are:

* `Node` always replaces the currently walked node, so return the parameter if you want nothing to change
* `Bool`, if true, continues to walk deeper, otherwise walking halts at that node.

I think given it is so new it is still safe to change. I wouldn't take this approach with other libraries but HCL is special given that I doubt people are using this API currently. Thoughts?

fatih:
@mitchellh This is all done :+1:  Please check the current code (I've force pushed it). Walk was used by some of the internal packages too, I've changed them and also add tests about various tasks, such as walking order, equality and rewrite.

mitchellh:
Looks great! Yes, as far as I know `Walk` is only used by internal packages so this is good to go!

PR #56, opened 2015-11-16
Adding Godoc link

derekdowling:
Makes understanding the consumable API much simpler rather than digging through source code.


mitchellh:
Legit, thanks.

PR #61, opened 2015-11-19
This fixes a bug in the flattener so that objects are flattened correctly

svanharmelen:
I guess we should add a better test for this, but this fixes the problem discussed here: https://github.com/hashicorp/terraform/issues/3964


jen20:
Thanks @svanharmelen, I also confirmed that this fixes the issue observed in Terraform.

PR #62, opened 2015-11-19
Restore old behavior for \" within interpolations

phinze:
Before the parser rewrite, HCL would silently convert `\"` within
interpolation braces to `"`.

After the conversion, this became a syntax error.

We've found several instances of Terraform configs in the wild using
this syntax. It results in a hard "syntax error" message during config
parsing. While avoiding the extra escape on double quotes within
interpolations is definitely preferred, the UX of the syntax error feels
harsh enough to be worth inserting this backwards compatibility for now,
leaving us the option of deprecating it with a warning down the line.

### DIFF
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -51,6 +51,13 @@ func TestDecode_interface(t *testing.T) {
 			},
 		},
 		{
+			"interpolate_escape.hcl",
+			false,
+			map[string]interface{}{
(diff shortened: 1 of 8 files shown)


mitchellh:
LGTM. I'm sad people were depending on this :( but fair enough.

phinze:
Yeah agreed on the sadness. C'est la vie!

PR #63, opened 2015-11-19
Handle multiple interpolations in string.

phinze:
A user testing an RC release of Terraform 0.6.7 found the following
string unexpectedly caused a syntax error:

    "echo ${var.region}${element(split(",",var.zones),0)}"

This fixes the error and covers it with a test.


mitchellh:
LGTM

PR #65, opened 2015-11-24
Fix scanning of identifiers of the form 'map.key'

jen20:
This is used in `.tfvars` files for Terraform in order to allow overriding values in maps, for example:

```hcl
    map.key1 = "Value"
    map.key2 = "OtherValue"
```

Not sure this is quite enough to fix the issues reported in 0.6.7 of Terraform, but it moves us past the parsing error.

### DIFF
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -44,6 +44,15 @@ func TestDecode_interface(t *testing.T) {
 			},
 		},
 		{
+			"tfvars.hcl",
+			false,
+			map[string]interface{}{
+				"regularvar": "Should work",
+				"map.key1":   "Value",
+				"map.key2":   "Other value",
+			},
+		},
+		{
 			"escape.hcl",
 			false,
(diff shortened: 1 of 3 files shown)


phinze:
LGTM

fatih:
Shouldn't we add tests for the scanner? It seems this is purely on how we define an identifier. So the tests should be in `scanner_test.go` rather than `decode_test.go`. Please correct me if I'm wrong :)

phinze:
@fatih good point! We're in "quick bugfix mode" to get a small terraform release out to address people's issues they're finding with 0.6.7, but you're totally right. If you're willing to PR, that'd be great, if you don't have the time at the moment, you can file it as an issue and we'll swing around to it. :+1:

fatih:
@phinze sounds god. No hurries, I'll check if I can send a PR if not I'll open an issue :)

PR #66, opened 2015-11-26
Support HEREDOCs in lists

jen20:
This fixes a regression in Terraform where HEREDOCS were previously supported in lists, reported in hashicorp/terraform#4065.


sethvargo:
:+1:

PR #68, opened 2015-11-28
Fixed hang when comment is on the last line of the file

Buzer:
Currently if the last line of the file contains a comment and there is no newline after, the scanner will hang indefinitely.


mitchellh:
This LGTM, thanks for the tests!

PR #71, opened 2015-12-01
Fix scanner for Windows line endings and add tests

jen20:
I think this should be enough to fix hashicorp/terraform#4069, though this is not yet verified and changes to the parser may also be necessary.

### DIFF
--- a/hcl/scanner/scanner.go
+++ b/hcl/scanner/scanner.go
@@ -399,6 +399,13 @@ func (s *Scanner) scanHeredoc() {
 		return
 	}
 
+	// Ignore the '\r' in Windows line endings
+	if ch == '\r' {
+		if s.peek() == '\n' {
+			ch = s.next()
+		}
+	}
+
 	// If we didn't reach a newline then that is also not good
 	if ch != '\n' {
 		s.err("invalid characters in heredoc anchor")
--- a/hcl/scanner/scanner_test.go
+++ b/hcl/scanner/scanner_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/hashicorp/hcl/hcl/token"
+	"strings"
 )
 
 var f100 = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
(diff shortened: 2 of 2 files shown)


hcl/scanner/scanner.go
```diff
@@ -224,7 +224,11 @@ func (s *Scanner) scanComment(ch rune) {
 	// single line comments
 	if ch == '#' || (ch == '/' && s.peek() != '*') {
 		ch = s.next()
-		for ch != '\n' && ch >= 0 && ch != eof {
+		for ch != '\n' && ch != '\r' && ch >= 0 && ch != eof {
+			if ch == '\r' && s.peek() == '\n' {
```
phinze: Confused about how if the guard has `&& ch != '\r'` this conditional will ever be true.

phinze: Oh sorry - by "the guard" I mean the one on L227.

phinze: In other words, I believe this will break comments immediately on `\r` instead of only on `\r\n`... which might be fine.

jen20: That's a typo (originally the whole guard was one line). Will fix.

phinze: :+1:

jen20: The question is do we want the `\r` to be included in the comment or treated as part of the line break? Unlike HEREDOCs I'd say they should be part of the line break for comments (though unless we're going to process comments it doesn't make much difference).

mitchellh:
On first pass something about this looks a little weird. 

The `ch != 'r'` test doesn't seem necessary in the for loop.

I didn't check the rest of the code but I believe due to how the scanner works the newline character(s) still show up in the Token text correct? I'm not sure if this is okay or not but it is worth questioning to make sure it is correct.

jen20:
@mitchellh The loop was a typo - now fixed. The current situation with this PR is:
- comments have the `\r` as part of the token text
- HEREDOCs keep the `\r` throughout (which is probably OK - unless we want to handle line ending conversion in e.g. Windows provisioners), but the `\r` can be part of the anchor so they terminate correctly.

phinze:
@jen20 and I both reviewed pre-rewrite behavior and confirmed that the behavior as he just described was is the same. this LGTM

PR #72, opened 2015-12-07
adds facility for deferred Decode similar to how json.RawMessage enables deferred Unmarshal

savaki:
decoded structs may now use the ast.Node type to save the raw ast.Node; conceptually similar to how json.RawMessage works and is useful in situations where you would like to either (a) defer decoding or (b) require additional meta-data like #Pos

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -15,6 +15,11 @@ import (
 // This is the tag to use with structures to have settings for HCL
 const tagName = "hcl"
 
+var (
+	// nodeType holds a reference to the type of ast.Node
+	nodeType reflect.Type = findNodeType()
+)
+
 // Decode reads the given input and decodes it into the structure
 // given by `out`.
 func Decode(out interface{}, in string) error {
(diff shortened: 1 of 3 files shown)


decoder.go
```diff
@@ -574,3 +587,12 @@ func (d *decoder) decodeStruct(name string, node ast.Node, result reflect.Value)
...
 	return nil
 }
+
+// findNodeType returns the type of ast.Node
+func findNodeType() reflect.Type {
+	var nodeContainer struct {
+		Node ast.Node
+	}
+	value := reflect.ValueOf(nodeContainer).FieldByName("Node")
+	return value.Type()
```
mitchellh: I think this can be simplified to `var v ast.Node; return reflect.ValueOf(v).Type()`, correct?

mitchellh:
Looks great! I made one little nitpick comment but otherwise looks good. I'll merge this week.

savaki:
Awesome.  Timing is perfect on this as I'm starting to use HCL pretty heavily for a commercial app that I'm developing.

PR #76, opened 2015-12-17
Fix crash on missing braces

phinze:
Fixes crash reported here https://github.com/hashicorp/terraform/issues/4350


mitchellh:
Not sure where the missing brace is but LGTM.

phinze:
```
resource "template_file" "cloud_config" {
  template = "$file("${path.module}/some/path")"
               ^         outer braces          ^
}
```

PR #77, opened 2015-12-18
add position to all decoder errors

phinze:
Several of the esoteric syntax errors we encounter in Terraform have
bubbled up errors from the decoder. Since all these errors have a Node
in context, they can report their position which makes them _much_ more
helpful to the user, even if the error message itself is confusing.

I tried to move around PosError somewhere but got stuck finding a good
spot for it that doesn't have either a silly name or an import cycle, so
for now I'm just cross referencing into the parser package to reference
the error type.

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/hcl/hcl/ast"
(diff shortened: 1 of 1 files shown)


mitchellh:
LGTM!

PR #80, opened 2016-01-06
Add new fmtcmd package (linting)

dcarley:
This uses the `printer` package to normalise HCL files. It's inspired by
github.com/fatih/hclfmt - but it differs in that it's intended to be more
generic, re-usable, and follow the semantics of `gofmt` or `go fmt`.

I intend to utilise this in Terraform by implementing `terraform fmt`
sub-command which will normalise all files in the current working directory
(like `go fmt ./..`) or contents over STDIN (with `terraform fmt -` that
could be used by editor plugins). So that Terraform users can benefit from
linting without installing another package/binary.

I hope that by placing most of the logic in the HCL package it should be
easy to implement sub-commands in other projects that also use HCL.

Some notes about the implementation:

- the significant difference from `gofmt` is that STDIN/STDOUT are passed in
  and errors aren't logged/written directly to STDERR, which gives consumers
  (e.g. Terraform) the ability to set appropriate exit codes
- I chose to use inline fixtures instead of files because there were a
  number of times where I needed to reference the contents and group them
  together with diff output
- it seemed simplest to construct the expected outputs by looping over the
  relevant fixtures and building up a string/byte slice, hope it isn't too
  confusing to read
- the test failure reporting is kind of rough because the outputs are so
  large, but I didn't want to add another diff function
- I chose to have a separate test for sub-directories rather than making it
  the default in the fixtures so that I didn't need to add additional logic
  to the fixture rendering
- the fixtures are sorted by filename before any of the tests runs so that
  they match the order that they are read from disk by `filepath.Walk()`

### DIFF
--- a/hcl/fmtcmd/fmtcmd.go
+++ b/hcl/fmtcmd/fmtcmd.go
@@ -1,0 +1,164 @@
+// Derivative work from:
+//	- https://golang.org/src/cmd/gofmt/gofmt.go
+//	- https://github.com/fatih/hclfmt
+
+package fmtcmd
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/hashicorp/hcl/hcl/printer"
+)
+
+var (
+	ErrWriteStdin = errors.New("cannot use write option with standard input")
+)
+
+type Options struct {
+	List  bool // list files whose formatting differs
+	Write bool // write result to (source) file instead of stdout
+	Diff  bool // display diffs instead of rewriting files
+}
+
+func isValidFile(f os.FileInfo, extensions []string) bool {
+	if !f.IsDir() && !strings.HasPrefix(f.Name(), ".") {
+		for _, ext := range extensions {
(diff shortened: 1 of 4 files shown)


dcarley:
Any thoughts on this?

One thing that occurred to me, while implementing tests for the Terraform part, was whether STDOUT was the most appropriate thing to pass around? It looks like it might result in copying lines/strings to `c.Ui.Output()`.

mitchellh:
Looks great! I think the prototype for this function is correct and we should look at how to implement this in Terraform and worry about the UI there. The writer can always be a bytes buffer or IO pipe we handle separately there.

mitchellh:
Sorry for the late response I totally missed this :)

dcarley:
Thanks! Terraform PR to follow soon..

echohack:
This isn't terribly important, but can you comment on why vertical whitespacing is used? @dcarley 

This is somewhat surprising to me because most languages I have seen either remove such whitespacing in their linters (python, json, rust) or have no opinion about it (ruby, go).

screen shot 2016-06-27 at 11 53 54

screen shot 2016-06-27 at 11 54 25

While I am a fan of having an automated standard for everyone to lint by, I'm not really a fan of hitting spacebar a ton. Maybe we can change this?

PR #84, opened 2016-02-07
Printer should not produce tailing spaces in lists

VladRassokhin:


VladRassokhin:
Also there's question about commas after last element. If all list elements is on same line, there no last comma. But if list is multiline comma after last element added. Is that intended?

mitchellh:
LGTM. Thanks for modifying the test! 

And yes, we want the comma on the last element.

PR #85, opened 2016-02-07
Fix indent of list item comments

VladRassokhin:
Previously comment for latest element would be indented incorrectly


mitchellh:
LGTM! Thanks!

PR #90, opened 2016-02-11
hcl/printer: group assignments into a single block

mitchellh:
This changes pretty printing so that assignments are grouped into a single block rather than always being separated by two newlines. You can see the test case for the example of what happens, but to make it easier:

```
foo = "bar"
longer = "baz"

other = "foo"

foo {
    foo = "bar"
    longer = "baz"
}
```

Goes to:

```
foo    = "bar"
longer = "baz"

other = "foo"

foo {
	foo    = "bar"
	longer = "baz"
}
```

Instead of:

```
foo    = "bar"

longer = "baz"

other = "foo"
```


mitchellh:
Going to try this a different way, the complexity is too high right now.

Issue #9, opened 2014-09-02
heredoc doesn't seem to handle tabs


dpetzel:
After reviewing https://github.com/hashicorp/hcl/issues/6, I attempted to implement heredoc for user_data in terraform. I had added the following:
```
    user_data = <<-USER_DATA

	#cloud-config

	coreos:
	  etcd:
		# multi-region and multi-cloud deployments need to use $public_ipv4
		addr: $private_ipv4:4001
		peer-addr: $private_ipv4:7001
	  units:
		- name: etcd.service
		  command: start
		- name: fleet.service
		  command: start

	USER_DATA
```
This throws the following error:
`Error loading config: Error parsing /my_file.tf: Line 76, column 0: syntax error`

It appears hcl is not honoring the `-` as outlined in http://en.wikipedia.org/wiki/Here_document:
> Appending a minus sign to the << has the effect that leading tabs are ignored. This allows indenting here documents in shell scripts without changing their value.

If I unindent everything and remote the `-` it works OK, but IMO the file becomes much harder to read since all the heredoc data needs to be completely left-aligned.

Phomias:
> In fact, I don't think - in any language ignores the leading tabs on the data in the heredoc, just the end marker. At the very least, Ruby is this way, but I tend to remember that being the case universally.

It's not quite universal. In BASH at least the leading tabs on the data are ignored when you use a `-` after the `<<` 

Try it yourself. To type a tab use `ctrl+v` `tab`

```
$ cat <<-EOF
>       this
>       that
>       the other
>       EOF
``` 
```
this
that
the other
``` 
```
$ cat <<-EOF
>       this
>       that
>       the other
> EOF
``` 
```
this
that
the other
``` 
```
$ cat <<EOF
>       this
>       that
>       the other
> EOF
```
```
	this
	that
	the other
```

I'd love it if you aim for the bash-like behavior because there are times (like with json policy documents on aws) when you can't have leading whitespace.  In fact, taking a [page out of the YAML book](https://en.wikipedia.org/wiki/YAML#Newlines_preserved), it would make me happiest (being a space-indenter) to have the leading (**spaces** or tabs) indent (of the first line) stripped from every line of the data.

Right now I have to use this:
```
resource "aws_iam_group_policy" "asset_managers_policy" {
  name =                        "asset_managers_policy"
  group = "${aws_iam_group.asset_managers.id}"
  policy = <<EOF
{
      "Version": "2012-10-17",
      "Statement": [
        {
[...]

PR #91, opened 2016-02-22
Add support for indented HEREDOC terminators

jen20:
This PR adds support for the style of HEREDOC often used in Ruby which allows the terminating marker to be indented if the HEREDOC is started with the sequence "<<-" rather than the usual "<<". This allows users to express documents with more natural indentation:

```
resource "template_file" "test" {
    template = <<-HEREDOC
        First Line
        Second Line
    HEREDOC
}
```

Note that this does not attempt to add any semantics around removing hanging indent from the actual text of the document, so extra indentation would still be present. We could make use of the canonical style for HCL herre to remove the hanging indent in the style of Ruby which would probably be more predictable for users.

Fixes https://github.com/hashicorp/hcl/issues/9

cc @phinze, @mitchellh

### DIFF
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -92,6 +92,11 @@ func TestDecode_interface(t *testing.T) {
 			map[string]interface{}{"foo": testhelper.Unix2dos("bar\nbaz\n")},
 		},
 		{
+			"multiline_indented.hcl",
+			false,
+			map[string]interface{}{"foo": "        bar\n        baz\n"},
+		},
+		{
 			"multiline_no_eof.hcl",
 			false,
 			map[string]interface{}{"foo": testhelper.Unix2dos("bar\nbaz\n"), "key": "value"},
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -94,7 +94,12 @@ func TestDecode_interface(t *testing.T) {
 		{
 			"multiline_indented.hcl",
 			false,
-			map[string]interface{}{"foo": "        bar\n        baz\n"},
+			map[string]interface{}{"foo": "  bar\n  baz\n"},
+		},
+		{
+			"multiline_no_hanging_indent.hcl",
+			false,
+			map[string]interface{}{"foo": "  baz\n    bar\n      foo\n"},
 		},
 		{
 			"multiline_no_eof.hcl",
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -94,12 +94,12 @@ func TestDecode_interface(t *testing.T) {
 		{
 			"multiline_indented.hcl",
 			false,
-			map[string]interface{}{"foo": "  bar\n  baz\n"},
+			map[string]interface{}{"foo": testhelper.Unix2dos("  bar\n  baz\n")},
 		},
 		{
 			"multiline_no_hanging_indent.hcl",
 			false,
-			map[string]interface{}{"foo": "  baz\n    bar\n      foo\n"},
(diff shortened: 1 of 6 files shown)


phinze:
Haven't reviewed the code yet, but +1 from me on the fuctionality, I have definitely missed this functionality from our HEREDOC implementation thus far.

I think removing the hanging indent is worth doing - it's a nicer experience for config authors.

mitchellh:
I super want this. I started adding this awhile back but stopped due to complications primarily with the printer package. I see you haven't gotten there yet but I predict it will be fairly annoying. 

I agree that if we do this I think we should automatically remove the equivalent indent from the text. That is what makes this really very useful. The rules around this are somewhat odd, but I think it should be:

* If all text has whitespace characters at least equal to the whitespace before the ending HEREDOC, trim it.

* If any text has whitespace less than the whitespace before the ending HEREDOC (i.e. the text itself is not indented), then ignore all lines, and don't trim _any_ whitespace.

I know other languages that do this, instead of going with above I would just see their rules and copy, probably. No need for us to reinvent the wheel.

jen20:
My plan for the whitespace was to copy what Ruby does - the rules you mention do sound sensible however. I haven't yet looked at the printer - but it should be valid in the first instance to output a regular unindented heredoc as semantically equivalent? When I come back to this (after Terraform 0.6.12) I'll do a roundup of popular implementations of this and post in here.

mitchellh:
That sounds like a great idea @jen20.

If I remember, the most clever language around heredocs that I actually liked was Scala. I think. Scala itself is a really complicated language but I remember the heredoc rules working as I expected and making the code look pretty. It may have been something else though...

jen20:
I've implemented a first pass at this using the semantics explained in the commit message and added some tests around expected behaviour. I think it matches the description @mitchellh gave in [this earlier comment](https://github.com/hashicorp/hcl/pull/91#issuecomment-187338389). Effectively:

- If your heredoc starts with `<<`, results are as before
- If your heredoc starts with `<<-`, and every line has the same prefix (tabs/spaces in the same order) as the terminating marker line, the prefix is removed from each line
- If your heredoc starts with `<<-`, but not every line has the prefix, the heredoc is returned in the same manner as if it specified with `<<`, but with the whitespace leading the marker removed.

This requires some testing on Windows - I am not sure of the effect of line endings.

phinze:
Behavior sounds good. Code looks reasonable to me. Windows testing does look like it will be importangt though.

sethvargo:
@jen20 @phinze one thing I just learned in bash land that might be helpful in Terraform too is the following. If you quote the delimiter, no interpolation takes place inside the heredoc. This is useful if you want to put bash in your heredoc. In the case of Terraform, I can envision a scenario where I want to dynamically write a configuration file that uses a similar syntax to Terraform:

```hcl
inline = <<"EOF"
${this.is.plain-text} and not interpreted as Terraform syntax
EOF
```

mitchellh:
@sethvargo I'd say thats a little too non-obvious. The `<<` and `<<-` are well mimicked syntactical notions. We also don't have any other way to make a string "raw" so I think we can skip that for me.

Windows would be important. As long as it behaves like the old heredocs then we're good. We should probably add tests for that.

jen20:
Right now it does behave like a regular here doc on Windows, but the tests fail because autocrlf screws up the expectation. I'll write some tests with the windows line endings hard coded in go code rather than in a separate HCL file.

jen20:
I think this is good to go now - we have green CI on both AppVeyor (Windows) and Travis.

phinze:
LGTM

PR #97, opened 2016-03-10
Add encoding API functions

### LINKED ISSUE #4: Wouldn't Parse(input []byte) be more typical? Can it work in streaming mode?
That's what you get from ioutil.ReadFile / ioutil.ReadAll.

Could hcl work without slurping the whole input into memory at once? Just a *bufio.Reader oughta be enough.

### LINKED ISSUE #20: Support Unmarshaler-ish interface
I might be completely missing something, but for my application I'm missing a way to have something like the Unmarshaler interface from encoding/json (without having to go through the "raw" object).

Just using Decode() does 90% of the work for me correctly, but I have a few fields where I want to convert the input before it goes into my data structure (for example "1d" to int(86400)).

sethvargo:
The std encoding format is Unmarshal([]byte, interface{}) error.
This naturally lends itself well to being passed readfile or exhausted
reader/buffer.

- Closes #4 
- Closes #92 
- Fixes #20 

This is just a rebase of 92 with authorship maintained

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -21,6 +21,17 @@ var (
 	nodeType reflect.Type = findNodeType()
 )
 
+// Unmarshal accepts a byte slice as input and writes the
+// data to the value pointed to by v.
+func Unmarshal(bs []byte, v interface{}) error {
+	root, err := parse(bs)
+	if err != nil {
+		return err
+	}
+
+	return DecodeObject(v, root)
+}
+
 // Decode reads the given input and decodes it into the structure
 // given by `out`.
 func Decode(out interface{}, in string) error {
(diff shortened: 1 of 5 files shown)


decoder.go
```diff
@@ -21,6 +21,17 @@ var (
 	nodeType reflect.Type = findNodeType()
 )
 
+// Unmarshal accepts a byte slice as input and writes the
+// data to the value pointed to by v.
+func Unmarshal(bs []byte, v interface{}) error {
```
jen20: Is there an argument here for reordering the parameters as in Decode below? Practically it is Decode which doesn't honour the convention but that is in pretty heavy use now.

jen20:
LGTM with one minor note!

decoder.go
mitchellh: Go has some odd conventions where `Unmarshal` is data first, but things like `io.Copy` are destination first. :-\

PR #102, opened 2016-03-14
Parse error on keys without values

mitchellh:
@dadgar found that HCL doesn't parse error on keys without a value (see the test cases). This fixes that.


sethvargo:
Ah. I actually noticed this as well, but thought it was the right behavior. LGTM

jen20:
This fixes several open issues in Terraform as well, I will go through to close them later this evening.

PR #105, opened 2016-03-21
Return an error if object keys are not strings

jen20:
This now gives an error instead of a panic when encountering configuration such as described in hashicorp/terraform#5740:

```
resource "aws" "web" {
  provider = "aws" {
    region = "us-west-2"
  }
}
```

We now return an error message - "hcl object keys must be a string" instead of crashing.

Fixes hashicorp/terraform#5740.

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -337,6 +337,14 @@ func (d *decoder) decodeMap(name string, node ast.Node, result reflect.Value) er
 			continue
 		}
 
+		// github.com/hashicorp/terraform/issue/5740
+		if len(item.Keys) == 0 {
+			return &parser.PosError{
+				Pos: node.Pos(),
+				Err: fmt.Errorf("%s: map must have string keys", name),
+			}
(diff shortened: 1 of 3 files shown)


phinze:
LGTM

fatih:
Is this an error on Parser/Scanner level? If the key can be only string may this is the wrong place? 

Fatih Arslan

phinze:
Ah good point @fatih - I think this behavior is correct, but stylistically it should probably go there. @jen20 does that make sense to you?

jen20:
Agreed - this fixes the immediate cause of the crash in hashicorp/terraform#5470, but a fix in the lexer or parser would be better overall and likely provider better error messages.

fatih:
I think the parser is the place. The lexer (scanner) only scans and gives a stream of tokens. String is a valid string. The representation whether a string should be come after another string is the concern of the parser.

PR #106, opened 2016-03-21
Temporarily ignore printer and fmt tests on Windows

jen20:
The red CI build on Windows is making it harder to process actual bugs - neither printer or fmt are used in any HC projects currently so ignoring the tests on AppVeyor/Windows seems reasonable for now. At some point they need fixing to account for line endings.


phinze:
LGTM - let's open an issue to cover that TODO.

PR #112, opened 2016-04-06
printer: output EOF newline in formatted hcl

phinze:
Noticed that the `hclfmt` tool (which uses hcl/printer) was chomping
the trailing newline from config files.

Here we switch to the more conventional newline-at-EOF output.

In the process of changing this, I realized that:

 * The printer_test.go build tag was wrong, so it wasn't running
 * The format() test helper was not exercising Format

So both of those fixes are included here

### DIFF
--- a/hcl/printer/printer.go
+++ b/hcl/printer/printer.go
@@ -60,5 +60,8 @@ func Format(src []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	// Add trailing newline to result
+	buf.WriteString("\n")
+
 	return buf.Bytes(), nil
 }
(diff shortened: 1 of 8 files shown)


phinze:
cc @jen20 for :eyes:

hcl/printer/printer.go
```diff
@@ -60,5 +60,8 @@ func Format(src []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	// Add trailing newline to result
+	buf.WriteString("\n")
```
sethvargo: I'm going to defer to @jen20 since he's spent more sadness in the arena than me, but I think this is going to break the printer on Windows :frowning:. I think we need to write `/r/n` or whatever their newline character is?

phinze: It's already broken on windows - we opt out of the tests there. I did a quick google search on the topic, and `\n` is what Println does, and windows deals with it. Hence, this.

jen20:
LGTM

PR #113, opened 2016-04-14
Allow printer to print single list file

grubernaut:
Currently the HCL Printer will print a file that contains a single
ast.ObjectList with two newlines after every ast.ObjectItem.

```
config-value = "value"

use = "theForce"

donut = true
```

This allows for the single list of ast.ObjectItem's to be printed with
only a single new line.

```
config-value = "value"
use = "theForce"
donut = true
```

ast.ObjectItem's which contain a ListType will still have two newlines
inbetween them, as expected.

### DIFF
--- a/hcl/printer/testdata/basic_list.golden
+++ b/hcl/printer/testdata/basic_list.golden
@@ -1,0 +1,4 @@
+listed_var = "tested var"
+bacon = "delicious"
+number = 3
+bool_var = true
--- a/hcl/printer/testdata/basic_list.golden
+++ b/hcl/printer/testdata/basic_list.golden
@@ -2,3 +2,12 @@ listed_var = "tested var"
 bacon = "delicious"
 number = 3
 bool_var = true
+
+# another
+baz = "qux"
+
+block {}
+
+block2 {}
+
+foo = "again"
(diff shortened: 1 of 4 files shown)


mitchellh:
Can we add a test to make sure that different groups separated by comments or blocks behave correctly? Jammed into a single test:

```
foo = "bar"
bar = "baz"

# another
baz = "qux"

block {}

block2 {}

foo = "again"
```

hcl/printer/testdata/basic_list.golden
```diff
@@ -0,0 +1,13 @@
+listed_var = "tested var"
+bacon = "delicious"
+number = 3
+bool_var = true
```
sethvargo: I don't think this is right - I think these equal signs should be aligned. 

Could you also add a test case for when the comment isn't on the left column:

```hcl
foo = [
  # comment
  "abc",

  # another comment
  "foo",
]
```

oryband:
Ping. Any chance to get this merged?

grubernaut:
Hey @oryband, we actually had an internal conversation about this PR yesterday, and have a plan moving forward. We should be able to provide a more complete solution to this issue fairly soon. Thanks!

mitchellh:
Closing this for bitrot :) The printer has improved greatly and I'm not sure if this is fixed but we're slowly fixing an issue one at a time.

PR #115, opened 2016-04-26
fmtcmd: Remove extra newline

phinze:
The fmtcmd tests weren't running before due to the same build tag
problem as #112, so didn't catch the fact that there was an extra
newline being added here, which is now doubled thanks to #112.

Fixing up the build tag reveals failing tests - removing the extra
newline fixes the tests.


phinze:
cc @jen20 for 👀

jen20:
LGTM

PR #117, opened 2016-04-28
hcl/fmtcmd: Clarify docs for Options.Diff

dcarley:
This description was originally copied straight from gofmt[0]. But as noted
in hashicorp/terraform#6343, the "instead of rewriting" suggests that it
disables the `Write` option, whereas it is possible to enable both.

[0]: https://golang.org/src/cmd/gofmt/gofmt.go#L31


mitchellh:
LGTM!

PR #120, opened 2016-05-05
Ignore field when tag is "-" when decoding structs

mirdhyn:
Like encoding/json each exported struct field becomes a member of the object unless the field's tag is "-"

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -517,6 +517,12 @@ func (d *decoder) decodeStruct(name string, node ast.Node, result reflect.Value)
 		structType := structVal.Type()
 		for i := 0; i < structType.NumField(); i++ {
 			fieldType := structType.Field(i)
+			tagParts := strings.Split(fieldType.Tag.Get(tagName), ",")
+
+			// Ignore fields with tag name "-"
+			if tagParts[0] == "-" {
+				continue
+			}
 
 			if fieldType.Anonymous {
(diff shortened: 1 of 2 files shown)


decoder.go
```diff
@@ -517,37 +517,39 @@ func (d *decoder) decodeStruct(name string, node ast.Node, result reflect.Value)
 		structType := structVal.Type()
 		for i := 0; i < structType.NumField(); i++ {
 			fieldType := structType.Field(i)
+			tagParts := strings.Split(fieldType.Tag.Get(tagName), ",")
```
sethvargo: Hi @mirdhyn 

Thank you for the PR. I think this might be a little cleaner if we reverse the logical operator here and just `continue` if the tag is `-`, for example:

```go
tagParts := strings.Split(fieldType.Tag.Get(tagName), ",")

// Ignore fields with tag name "-"
if tagParts[0] == "-" {
  continue
}

if fieldType.Anonymous {
 // ...
}
```

This will save us the additional layer of indentation, and I think it has the same result. What do you think?

mirdhyn: I agree. I updated the PR.

sethvargo:
Great, thanks @mirdhyn 

Gonna defer to @jen20 for merge.

mitchellh:
LGTM!

PR #121, opened 2016-05-11
Skip a nil item during filtering

vishalnayak:
While testing vault-secure-intro, the config file had an undesired `=` in the config, which was causing a panic.

Wrong input causing panic:
environment = "aws" {
}

Correct input:
environment "aws" {
}

This is where the investigation lead to.
All the tests pass.

### DIFF
--- a/hcl/ast/ast.go
+++ b/hcl/ast/ast.go
@@ -56,7 +56,7 @@ func (o *ObjectList) Filter(keys ...string) *ObjectList {
 	var result ObjectList
 	for _, item := range o.Items {
 		// If there aren't enough keys, then ignore this
-		if len(item.Keys) < len(keys) {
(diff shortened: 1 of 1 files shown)


jen20:
:+1: good catch. Sorry this took so long to pick up, HCL has been off my radar for a while!

mitchellh:
Can we add a test for this?

mitchellh:
I added a test and this doesn't actually panic (it gives a syntax error now). @jen20 I actually think we shouldn't commit this because from what I've found a panic here is indicative of a higher-level logic issue (i.e. I recently found this one: https://github.com/hashicorp/hcl/commit/40a9504600303e80f0b0ef13b0b7e9779209ed98 which manifested itself as this panic).

jen20:
Reasonable - I'll revert this then.

PR #125, opened 2016-06-06
strconv: Fix escaped backslashes \\ in braces ${}

phinze:
Fixes hashicorp/terraform#6359 and several
related issues by unescaping a double backslash within braces to a
single backslash.

This bumps into the bit of HIL that still hangs out in HCL - for values
that aren't interpolated, like Terraform's variable defaults - users
were unable to get a backslash to show up within `${}`.

That's because `${}` is handled specially to allow for, e.g., double
quotes inside of braces.

Here, we add `\\` as a special cased escape (along with `\"`) so users
can get backslashes in these scenarios by doubling them up.

### DIFF
--- a/hcl/strconv/quote.go
+++ b/hcl/strconv/quote.go
@@ -49,7 +49,8 @@ func Unquote(s string) (t string, err error) {
 	for len(s) > 0 {
 		// If we're starting a '${}' then let it through un-unquoted.
 		// Specifically: we don't unquote any characters within the `${}`
-		// section, except for escaped quotes, which we handle specifically.
+		// section, except for escaped quotes and escaped backslashes, which we
+		// handle specifically.
 		if s[0] == '$' && len(s) > 1 && s[1] == '{' {
 			buf = append(buf, '$', '{')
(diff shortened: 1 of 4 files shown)


sethvargo:
👍

hcl/strconv/quote.go
```diff
@@ -70,6 +71,9 @@ func Unquote(s string) (t string, err error) {
 					if q, _ := utf8.DecodeRuneInString(s); q == '"' {
 						continue
 					}
+					if q, _ := utf8.DecodeRuneInString(s); q == '\\' {
+						continue
+					}
```
mitchellh: The performance is unlikely to matter but decoding utf8 runes is not exactly cheap, so perhaps we should just turn this into a `switch`?

phinze: Fair - I was just so punch drunk that the copy paste seemed too hilarious not to do 😄

phinze:
Now, with more `switch`!

mitchellh:
LGTM!

PR #132, opened 2016-06-16
Remove BC escaped double quote behavior

phinze:
When we switched to the current edition of the HCL parser, we
inadvertently broke some undocumented behavior where users could
successfully used escaped double quotes in a nested interpolation like
this:

    foo = "${somefunc(\"somearg\")}"

The proper syntax here is:

    foo = "${somefunc("somearg")}"

Because once you are inside the interpolation braces, the "quoting
context" is different.

At the time, while we didn't like the fact that the non-standard syntax
was in the wild, we didn't want to break BC, so we treated it as a bug
and fixed it in #62.

Now that we are at the moment of a major Terraform release, we can yank
the BC compatible behavior, which fixes a currently broken use case -
there is no way to express strings with double quotes in them from
inside interpolation braces.

    foo = "${somefunc("\"inquotes\"")}"

After merge and a dep upgrade, this will take care of
https://github.com/hashicorp/terraform/issues/5550


phinze:
cc @jen20 or @jbardin for review

mitchellh:
This seems reasonable to me. :+1:

hcl/strconv/quote.go
```diff
@@ -70,7 +70,7 @@ func Unquote(s string) (t string, err error) {
 				if r == '\\' {
 					q, _ := utf8.DecodeRuneInString(s)
 					switch q {
-					case '"', '\\':
```
jbardin: leftover comment above about leaving escaped double quotes

phinze:
Forgot about @mitchellh :smile:

Good catch @jbardin! Repushed w/ fixed comments - PTAL

jbardin:
LGTM

PR #133, opened 2016-06-20
Fix bug with unquoting null from JSON-encoded HCL

jen20:
Fixes hashicorp/terraform#6774.

### DIFF
--- a/hcl/token/token.go
+++ b/hcl/token/token.go
@@ -152,6 +152,11 @@ func (t Token) Value() interface{} {
 			f = strconv.Unquote
 		}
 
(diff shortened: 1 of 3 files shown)


hcl/token/token.go
```diff
@@ -152,6 +152,10 @@ func (t Token) Value() interface{} {
 			f = strconv.Unquote
 		}
 
+		if t.Text == "" {
```
mitchellh: Might help to have a comment here about when this could happen.

mitchellh:
Small comment but LGMT

Issue #128, opened 2016-06-07
Input Causes Panic: {:{


BrianHicks:
I've found a crashing input in HCL as of d7400db7143f8e869812e50a53acd6c8d92af3b8. Here's an [SSCCE](http://www.sscce.org/) demonstrating the problem.

```go
package main

import (
	"github.com/hashicorp/hcl"
)

var (
	data = 	"{:{"
)

func main() {
	var out interface{}
	hcl.Unmarshal([]byte(data), &out)
}
```

When run, this gives the following output:

```
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0x1ced60, 0xc82000a0b0)
	/var/folders/f6/l26tk3xs5pjczcd5k5pcrmk00000gn/T/go-fuzz-build846495901/src/runtime/panic.go:481 +0x3e6
github.com/hashicorp/hcl/hcl/ast.(*ObjectList).Pos(0xc820010200, 0x0, 0xc8200413b0, 0xc8200413e0, 0xc8200413b0, 0x15)
	/var/folders/f6/l26tk3xs5pjczcd5k5pcrmk00000gn/T/go-fuzz-build846495901/src/github.com/hashicorp/hcl/hcl/ast/ast.go:112 +0x11f
github.com/hashicorp/hcl.(*decoder).decodeMap(0xc820041d70, 0x1fc790, 0x4, 0x383058, 0xc820010200, 0x19b160, 0xc820012150, 0x15, 0x0, 0x0)
	/var/folders/f6/l26tk3xs5pjczcd5k5pcrmk00000gn/T/go-fuzz-build846495901/src/github.com/hashicorp/hcl/decoder.go:343 +0x1083
github.com/hashicorp/hcl.(*decoder).decode(0xc820041d70, 0x1fc790, 0x4, 0x383058, 0xc820010200, 0x19a5a0, 0xc82000a120, 0x194, 0x0, 0x0)
	/var/folders/f6/l26tk3xs5pjczcd5k5pcrmk00000gn/T/go-fuzz-build846495901/src/github.com/hashicorp/hcl/decoder.go:100 +0xcc6
github.com/hashicorp/hcl.(*decoder).decodeInterface(0xc820041d70, 0x1fc790, 0x4, 0x383058, 0xc820010200, 0x19a5a0, 0xc82000a120, 0x194, 0x0, 0x0)
[...]

PR #134, opened 2016-06-21
Fix parser bugs found via fuzzing from GH-128

mitchellh:
This fixes all the inputs found in GH-128 to no longer panic.

There are some legitimate edge cases found here, and others that are jus very unlikely to be ever hit. :)


jen20:
LGTM!

PR #137, opened 2016-07-06
hcl/parser: Support lists of objects

phinze:
a.k.a lists of maps

Implementation was pretty straightforward - I had to tweak the `needsComma`
handling since it was stuck inside literal parsing. It happens out front
now. I also promoted the `assign_deep.hcl` parser test to a decoder
test that passes, since it was testing for an error to occur but now it
works! :)

Additionally we make ObjectLists support being comma-delimited, which
enables maps to defined inline like `{one = 1, two = 2}`.

Refs https://github.com/hashicorp/terraform/issues/7142


phinze:
cc @jbardin, @jen20, @mitchellh

VladRassokhin:
Please also change `syntax` section in `README.md`

mitchellh:
The parser section looks good to me, but I'd also take a look at the printer package to make sure that handles these in a nice way.

phinze:
Good call @VladRassokhin on the docs update, and thanks @jen20 for picking up the printer test and docs tweaks here. This all LGTM now.

PR #138, opened 2016-07-11
Support multiline string literals in HCL

jen20:
This allows multiline strings to be parsed in order to make HIL interpolations nicer in Terraform:

```
my_complex_thing = "${merge(var.list1,
                            var.list2,
                            var.list3)}"
```

cc @mitchellh @phinze


phinze:
This LGTM

VladRassokhin:
So now both string literals (single or double quoted) and heredocs supports multiline strings? Or only HIL interpolation could be multiline and evaluated content of quoted string should not contain EOL?

This language becomes weirder^W more complex every day.

PR #139, opened 2016-07-22
Support EOL in string literals only in HIL fragments

VladRassokhin:
Alters #138, so only interpolations may contain `\n`.
If user want multiline string, heredoc should be used.


VladRassokhin:
@jen20 @phinze Guys, please review.
From IDE point of view, allowing string literals to be multiline could lead to some huge reparses. Multiline string is like non-terminated `/*` comment in other languages: it's hard to detect where it started and where it should have ended.
In current (#138) solution once user typed `"` everything until next `"` would be lexed as string, and same for next quote and so on, so almost all document would be relexed and reparsed.

jen20:
Hi @VladRassokhin! Sorry this has taken some time to get to. I think I agree that this is the correct resolution - multiline variables are not allowed in HIL (currently). The only thing I am somewhat concerned about is that this muddies the waters between HCL and HIL (currently independent). Going to defer this one to @mitchellh and/or @phinze here.

mitchellh:
I agree but I think we have to merge this to coincide with Terraform 0.8 otherwise we'll introduce BC in the middle of a TF release cycle. There are other projects that depend on this too but I can't think of any where its likely that this would cause issues...

Its unfortunate but hopefully we're getting to a point now where this is becoming much rarer in HCL.

mkuzmin:
Now Terraform 0.8 went into beta phase, that's a good time to merge this PR in.

mkuzmin:
@mitchellh is there a chance to have this merged in Terraform 0.8?

This is not just an enhancement. Since version 0.7 auto-completion for HCL code in IntelliJ IDE does not work correctly.

VladRassokhin:
I've rebased branch to current master.
@mitchellh, @jen20 merge it before Terraform 0.8, please.

mitchellh:
Agreed. Merging.

PR #143, opened 2016-08-22
Fix panic when decoding invalid value structure into struct

mitchellh:
Fixes https://github.com/hashicorp/terraform/issues/8295

When decoding an object into a struct where the object structure doesn't
match the Go struct structure, the case tested here would panic. This
introduces additional checks to guard against the edge case being hit to
avoid the panic.

The specific checks being added are: if an item being decoded into a
struct is a literal type, the item to be decoded must be non-nil in
order to use it. This isn't super clear and to be honest I also don't
fully understand it but this fixes the problem without introducing any
more test failures and without significant code complexity.


sethvargo:
LGTM

PR #145, opened 2016-09-01
json: interpolations have to be escaped

mitchellh:
Related: https://github.com/hashicorp/terraform/issues/8110

At some point we ignored the " in interpolations. We do this for HCL and
it is correct but this is invalid JSON syntax and for JSON we've always
had the stance that we have to escape them.

This shows up in HCL as normal unescaped (because it only escapes it for the purpose of JSON), as the test shows in decode_test.go.


sethvargo:
👍

PR #146, opened 2016-09-02
hcl/scanner: single line '//' commments verify second '/'

mitchellh:
Fixes: https://github.com/hashicorp/terraform/issues/8477

The way the scanner works '/ foo' was actually valid comment syntax.
This obviously is not what we want. This modifies the scanner to verify
that '//' comments in fact have the second '/'.


jen20:
:+1:

PR #149, opened 2016-09-05
json/parser: set Assign property when parsing object item

Acconut:
Without this patch the `json/parser.Parser` did not fill the `hcl/ast.ObjectItem.Assign` property resulting in missing assignment operators, visible when printing it using the `hcl/printer` package. Please look at following example code demonstrating the issue:

```go
package main

import (
  "os"

  "github.com/hashicorp/hcl/json/parser"
  "github.com/hashicorp/hcl/hcl/printer"
)

const input = `
{
  "a": "b"
}
`

func main() {
  ast, err := parser.Parse([]byte(input))
  if err != nil {
    panic(err)
  }

  err = printer.Fprint(os.Stdout, ast)
  if err != nil {
    panic(err)
  }
}

```

When executed, this program printed `"a" "b"` to stdout which is missing the `=` character.
With the patch applied, the output is the correct version: `"a" = "b"`

### DIFF
--- a/json/parser/parser.go
+++ b/json/parser/parser.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/hcl/hcl/ast"
+	hcltoken "github.com/hashicorp/hcl/hcl/token"
 	"github.com/hashicorp/hcl/json/scanner"
 	"github.com/hashicorp/hcl/json/token"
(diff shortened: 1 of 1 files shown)


sethvargo:
👍

Acconut:
Thank you for merging!

PR #150, opened 2016-09-10
[DO NOT MERGE YET] hcl: don't escape ANYTHING in ${}, let lower layer handle it

mitchellh:
🚧  **DO NOT MERGE YET**  🚧 

This changes the behavior of HCL to _completely ignore any escape sequences within `${}`_. This is likely to break some configurations that used this. However, it resulted in more broken behavior if used with HIL. 

The prior behavior caused problems because HCL would escape some things (escaped backslashes) and then HIL would do the same, or error for syntax/unescaped reasons. This double escape caused most practical configs to not work at all.

This change is slated for Terraform 0.8 when we can introduce BC-incompat.

### DIFF
--- a/decoder_test.go
+++ b/decoder_test.go
@@ -63,7 +63,7 @@ func TestDecode_interface(t *testing.T) {
 				"qux":          "back\\slash",
 				"bar":          "new\nline",
 				"qax":          `slash\:colon`,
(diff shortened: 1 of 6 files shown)


mitchellh:
Terraform 0.8 is upon us. Here we go.

PR #152, opened 2016-09-13
missing fields when decoding JSON

jbardin:
A single JSON object with a single key and value of a another object fails to decode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,3 @@
 sudo: false
 language: go
-go: 1.5
+go: 1.7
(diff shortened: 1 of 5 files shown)


jbardin:
This will fix #153

jbardin:
back to WIP -- this doesn't solve the terraform issue, so we need to expand maps in some cases too.

jbardin:
All tests green! 🍏

mitchellh:
Looks really great! :D

jbardin:
rebased in some spelling/comment fixes.

PR #163, opened 2016-10-31
Add failing format example

sethvargo:
This is still blocking Terraform fmt training 😦

### DIFF
--- a/hcl/printer/nodes.go
+++ b/hcl/printer/nodes.go
@@ -118,17 +118,27 @@ func (p *printer) output(n interface{}) []byte {
 		defer un(trace(p, "ObjectList"))
 
 		var index int
-		var nextItem token.Pos
 		for {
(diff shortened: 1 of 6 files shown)


VladRassokhin:
Looking at all comments-related issues, maybe it's worth to support comments as standalone nodes in AST? That would simplify all 'pre-comment' 'post-comment' nodes for elements in AST.

You could just add rules how related (siblings, parent/child) AST nodes should be positioned and all comments would be formatted according that rules.
Inspired by how it's done in IntelliJ platform where comments are separate nodes AST and PSI(another tree over AST). Also there's special 'Formatting' tree over PSI that's processed with formatting rules.

mitchellh:
@VladRassokhin Yeah that's not a bad idea at all! Would be a bigger refactor but I'm certainly not against it.

PR #169, opened 2016-11-12
Remove duplicate unit test

jMarkP:
Tiny change! Two unit test cases were testing parsing of the same list of literals.


mitchellh:
Thanks :)

PR #171, opened 2016-11-24
Integrate HIL parsing into the decoder

apparentlymart:
Previously the integration between HCL and HIL was very loose: HCL made some accommodations in its string parsing so that HIL's `${ ... }` sequences got passed through without any extra escaping, but otherwise ignored it.

Treating HIL as an entirely separate concern from HCL generally causes applications to decode the raw HIL expression source into a string variable and then pass it to the HIL parser, but then the HIL parser
doesn't get the context about where in the HCL source the expression is located.

To make it easier for calling applications to "do the right thing" when accepting HIL expressions from HCL, here we deepen slightly the integration such that a caller can provide a HIL `ast.Node`
and the HCL decoder will then automatically pass a corresponding string in the configuration to the HIL parser.

We also support literals of other primitive types and just return them as synthetic HIL literals, so that callers don't need to bend over backwards to handle the common case where a value can either be a literal primitive or a HIL expression.

This direct coupling between HCL and HIL represents an architectural change, but one that acknowledges that the primary use-case for HIL is being embedded in HCL, and makes it easier for calling applications to correctly report errors with HIL-in-HCL with the appropriate source position.

This change requires hashicorp/hil#37 .

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -11,6 +11,9 @@ import (
 	"github.com/hashicorp/hcl/hcl/ast"
 	"github.com/hashicorp/hcl/hcl/parser"
 	"github.com/hashicorp/hcl/hcl/token"
+	"github.com/hashicorp/hil"
+	hilAST "github.com/hashicorp/hil/ast"
+	hilParser "github.com/hashicorp/hil/parser"
 )
 
 // This is the tag to use with structures to have settings for HCL
@@ -19,6 +22,9 @@ const tagName = "hcl"
 var (
 	// nodeType holds a reference to the type of ast.Node
 	nodeType reflect.Type = findNodeType()
+
+	// hilNodeType holds a reference to the type of hilAST.Node
+	hilNodeType reflect.Type = findHILNodeType()
 )
 
 // Unmarshal accepts a byte slice as input and writes the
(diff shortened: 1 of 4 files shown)


apparentlymart:
@mitchellh this is now updated to work with the changes to hashicorp/hil#37.

This change doesn't really do much on its own. My thought here was that this would be one small step towards making the other HCL/HIL-using tools (Terraform being my primary concern, of course!) be able to easily produce good error messages with proper source positions when HIL parsing (and later, evaluation) fails.

Whether Terraform itself would use this exact mechanism is debatable, since its usage of HCL/HIL is more "deep" than some of the other callers. One possibility I'd considered, which I'd be interested to discuss either here or elsewhere, is for a future Terraform major release to change `config.RawConfig` to contain HIL nodes rather than strings, in which case HIL parsing would happen during configuration parsing rather than validation/interpolation, and could thus potentially use the mechanism added by this commit.

jippi:
bump!

Xuanwo:
hcl v2 has been released, may this issue can be closed?

apparentlymart:
Oh yes, thank you for pointing that out @Xuanwo!

For anyone else seeing this and wondering what happened: Integrating HIL with HCL (beyond what I started here) turned out to be pretty impractical without breaking API changes, and so this initial idea is what developed into HCL 2.0, now available as [`github.com/hashicorp/hcl/v2`](https://pkg.go.dev/github.com/hashicorp/hcl/v2). HCL 2 incorporates a HIL-like expression language, effectively merging the two into a single language so that HIL as a separate entity is no longer needed.

Issue #107, opened 2016-03-21
Fix printer and fmt tests on Windows


jen20:
See #106 for more context.

There are some failing tests which I have tentatively attributed to line endings on Windows. However, these test files are currently ignored on Windows as they are not used by any HashiCorp products right now.

PR #173, opened 2016-11-30
Remove CRLF line endings

jbardin:
Normalize HCL input to use "\n" line endings. The scanner and output
only work with the '\n' character, and "\r\n" newlines can leave
dangling "\r" characters in the parsed output.


jbardin:
This should finish up #107, and also takes care of terraform stripping most windows newlines reported in terraform#10381.

mitchellh:
Hm, so in cases like this I usually try to see what other tools do. The closest in this case is of course `go fmt`. `go fmt` enforces `\n` as the line terminator, and it makes sense our printer works the same way since its based off that. I'd say thats the behavior we should have.

I checked to see what happens with `\r\n` even in raw strings in Go and they're force converted to `\n` by `fmt`, so I think thats fine behavior. Basically, if a user wants explicit \r\n, they need to use a string with "\r\n" explicitly (literally). 

I'll change this slightly since I think the only thing to do here is make sure the output doesn't have \r\n.

mitchellh:
Actually I don't have access very quickly to a Windows machine to run tests atm, at least for the next few days. @jbardin do you mind doing the above? ^

jbardin:
SGTM! That make the change even simpler.

jbardin:
@mitchellh: I forgot about this, but it would be nice to have a single official file format for TF configs. Issues from windows line ending crop up every so often.

tintoy:
Hi - stupid question, but if `\r\n` is simply replaced with `\n`, won't the positions in the parsed AST be incorrect with respect to the original file contents?

The reason that I ask is I was looking into building an LSP language service for Terraform and would rather avoid writing my own HCL parser if necessary. But if the positional information in the AST cannot be relied on when there are Windows-style (or mixed) line endings then that could be tricky.

PR #174, opened 2016-12-01
Try to use SetInt in case if possible

kron4eg:
In my config file I have many time.Duration fields, which were decoded perfectly fine with JSON, but with HCL it panicked, so here is the fix that will allow to decode aliased int types, like time.Duration.


mitchellh:
LGTM.

PR #176, opened 2016-12-15
Modify new bytes.ContainsRune to backward compatible bytes.IndexRune

theherk:
The purpose of this change is so that the code is valid in Go < 1.7. I am using hcl/printer in spf13/viper#287. It requires tests in Go < 1.7, so this prevents that from passage. I did a work around using build constraints, but it is pretty ugly. It would be much better without that workaround being required and there will be no impact on hcl.


mitchellh:
In general we only support the latest Go version and ask you use vendoring to support prior versions. But due to the simplicity of this change I'll take it. :)

theherk:
That's great to hear @mitchellh. Thank you for your understanding.

PR #195, opened 2017-04-17
Document autotrim of whitespace with hyphen in heredoc

syl20bnr:
This form is used in the new book `Terraform: Up & Running` but it seems that it is not documented.
This PR mention this hyphen form in the README.

Would also need to be documented in Terraform documentation: https://www.terraform.io/docs/configuration/syntax.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,12 +66,24 @@ of the syntax and grammar is listed here.
 
   * Multi-line strings start with `<<EOF` at the end of a line, and end
     with `EOF` on its own line ([here documents](https://en.wikipedia.org/wiki/Here_document)).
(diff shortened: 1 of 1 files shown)


mdaniel:
Ironically, I was just making that change to the file in `terraform-website` and came here looking for some supporting documentation I could link to. I now know why I didn't find it. :-)

There [doesn't seem](https://github.com/hashicorp/terraform-website/issues?utf8=%E2%9C%93&q=is%3Aissue+heredoc) to be an issue for this change in that repo; since there is one here, should I go ahead and put the website change in PR, or wait for this issue to resolve first in order to have a better language template for the TF side?

PR #197, opened 2017-04-20
Return an error instead of printing stray `illegal\n` to stdout.

sean-:
Encountered while running `hcl.Decode()` on a JSON blob.


mitchellh:
Hah! Thanks, yeah there are some jank things like this lingering around. Thanks. :)

Issue #99, opened 2016-03-12
Custom marshalling support


let4be:
When working with JSON I can simply define UnmarshalJSON to define my custom format(like parsing string encoded `time.Duration`)
Is there any easy way to do the same using HCL?
Thanks!

let4be:
Oh, I see there is already a request for https://github.com/hashicorp/hcl/issues/89

Issue #89, opened 2016-02-11
Feature request: support for TextUnmarshaler


dim13:
With TextUnmarshaller own data types can be defined, such as a wrapper for time.Duration, etc.

PR #203, opened 2017-06-10
Support decoding HCL via an Unmarshaler or TextUnmarshaler

nilium:
Unmarshaler is a basic interface that any type can implement. It allows
the type being unmarshaled over to control how it's unmarshaled --
e.g., bit flags could be unmarshaled from a list and so on.

encoding.TextUnmarshaler supports a smaller range of literal token
types, namely only strings, heredocs, floats, numbers, and bools.
Unlike decoding an ast.Node into a string using DecodeObject, this will
use the text of floats and bools (whereas for DecodeObject, it's an
error to decode a float or bool into a string).

Related to #20, #89, #99, and #168 (the last of which I decided not to
go through due to the number of changes / tests failing).

### DIFF
--- a/decoder.go
+++ b/decoder.go
@@ -1,6 +1,7 @@
 package hcl
 
 import (
+	"encoding"
 	"errors"
 	"fmt"
 	"reflect"
@@ -12,6 +13,13 @@ import (
 	"github.com/hashicorp/hcl/hcl/parser"
 	"github.com/hashicorp/hcl/hcl/token"
 )
+
+// Unmarshaler is the interface implemented by types that wish to override the unmarshalling
+// behavior of hcl. The input to the UnmarshalHCL function is the ast.Node associated with the value
+// being unmarshaled.
+type Unmarshaler interface {
+	UnmarshalHCL(node ast.Node) error
+}
 
 // This is the tag to use with structures to have settings for HCL
 const tagName = "hcl"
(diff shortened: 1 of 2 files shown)


nilium:
I want to point out this is still in need of further review / thinking. Particularly the `(*decoder).tryDecodeUnmarshaler` method, since it involves reflection, isn't my strongest area, and involves patching code using reflect already.

nilium:
I wonder if the Unmarshaler's UnmarshalHCL method should accept the name of the node being unmarshaled? E.g.,

```go
type Unmarshaler interface {
        UnmarshalHCL(name string, node ast.Node) error
}
```

This would allow it to make possibly better decisions about how to react to the node, but I'm also not sure I can come up with a really good example scenario for that either. Figured I'd throw that out there, at any rate.

nilium:
Any hcl maintainers had a chance to look at this? Been about a month now.

nilium:
I'm guessing this isn't going to happen? I'd appreciate some feedback from a maintainer so I can know whether I should close this and just maintain a fork of hcl with these patches -- that's not a problem to do, either, just don't want to leave this hanging as it is.

elewis787:
I am guessing you are correct. I recently found https://github.com/hashicorp/hcl2 . 

Any roadmap docs or plans for hcl that the maintainers could share ?

nilium:
Closing this due to inactivity. No point keeping it open since it appears hcl is only accepting bug fixes now.

PR #206, opened 2017-07-30
[WIP] printer: Add support for external filters during Format

vancluever:
This PR introduces the ability to pass in external filters to alter the
formatting process in a way specific to the tool that HCL is implemented
for. An example would be pre-populating fields in Terraform during
`terraform fmt`, such as the description fields for variables and outputs.

`printer.Filter` is specifically an interface - the downstream tool
supplies the `Filter` method, which carries out the necessary behaviour.

Also in this PR are some updates to docs for `Run` and `Options` in `fmtcmd`.

### DIFF
--- a/hcl/fmtcmd/fmtcmd.go
+++ b/hcl/fmtcmd/fmtcmd.go
@@ -22,10 +22,13 @@ var (
 	ErrWriteStdin = errors.New("cannot use write option with standard input")
 )
 
+// Options are options that can be passed to Run to affect the behaviour of the
+// formatter.
 type Options struct {
-	List  bool // list files whose formatting differs
-	Write bool // write result to (source) file instead of stdout
-	Diff  bool // display diffs of formatting changes
+	List    bool             // list files whose formatting differs
+	Write   bool             // write result to (source) file instead of stdout
+	Diff    bool             // display diffs of formatting changes
+	Filters []printer.Filter // Any externally-supplied filters for the printer
 }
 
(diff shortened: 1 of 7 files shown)


VladRassokhin:
AFAIU there would be a problems once printer would try to restore position of comments.
So for now Filters are safe unless they don't change size of elements.

Could you please add complex test case with filter altering/inserting AST nodes on a file with comments?

vancluever:
Hey Vlad, will do. I am working on this specific case actually for the updates that I'm making to `terraform fmt`, for which adding nodes was my specific use case - I'll port some of those tests over to HCL when I'm done over there.

vancluever:
Hey @VladRassokhin - just WIPing this as there are a couple of things: First off I'm seeing some formatting artifacts in the related TF PR (referenced in the mentions) that I think once they are fixed should be included as a test/example in HCL as well. The issue is not directly related to the HCL part of the feature, as the HCL part really only supplies the interface, but I think it's an important PoC. (I think this might be what you were alluding to - things are fine, as long as the file doesn't have standalone comments.)

Second off is that @apparentlymart and I are talking about the details of the implementation in Terraform specifically and it could possibly mean that whatever that ends up looking like means that this might not be necessary, and I don't necessarily want to merge it if it doesn't have a strong use case.

Will come back to this soon. Sorry for jumping the gun!

vancluever:
Hey all - closing this as I think I want to re-think my approach on this one. Will re-submit once I have a bit more of an idea of what this looks like. Sorry for the delay in updating!

PR #230, opened 2018-01-03
Consider empty lines as indented

jocgir:
I just discovered that it is possible to automatically unindent heredoc (even if it is not yet in the documentation #195). Nice feature.

However, if I declare something like that:
```
description = <<-EOF
    This is the fist line

    This is the second line
    EOF
```

It may not work because the empty line between the two lines may have its trailing blanks removed by the editor. It works only if the line starts with spaces.

To fix that problem, I suggest that you do not consider empty lines as unindented lines to avoid inconsistent behaviour.

### DIFF
--- a/hcl/token/token.go
+++ b/hcl/token/token.go
@@ -191,7 +191,7 @@ func unindentHeredoc(heredoc string) string {
 
 	isIndented := true
 	for _, v := range lines {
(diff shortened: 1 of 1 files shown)


jocgir:
Same as #259, but since @julian7 added tests, his is better :-).

I know that there is an hcl2, but it would be nice if this one could be integrated.

PR #241, opened 2018-03-20
printer, scanner: Don't produce unparsable output.

octo:
This changes fix two cases in which `printer.Format()` produced output that was not accepted by `parser.Parse()`:

1.  The input ends in `\x00`: the scanner accepted this, because nothing is following the null byte. The formatter, however, decides to add a newline at the end of the file, assuming that this would be a no-op. Upon re-parsing, the null byte is no longer the last character in the stream and produces an error.
1.  The input contains the U+E123 unicode codepoint: this codepoint is used internally by the formatter to do some postprocessing of the generated output. If the input contains that codepoint, the postprocessing removes semantically relevant information, for example the pound sign signifying the beginning of a comment. This leads to unparsable output.

I think that it's sane to refuse those inputs as "invalid" and I think it's very unlikely that this will break existing use-cases.

### DIFF
--- a/hcl/printer/printer_test.go
+++ b/hcl/printer/printer_test.go
@@ -147,3 +147,25 @@ func lineAt(text []byte, offs int) []byte {
 	}
 	return text[offs:i]
 }
+
+// TestFormatParsable ensures that the output of Format() is can be parsed again.
+func TestFormatValidOutput(t *testing.T) {
+	cases := []string{
+		"#\x00",
+		"#\ue123t",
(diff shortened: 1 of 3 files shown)


mitchellh:
Thanks, this looks good. Note on point #2 that technically we probably should disallow any of the private use code points. I chose U+E123 because its slightly hinted at being human-made (1-2-3), but there are ~18,000 private use code points (HCL only uses one). 

Thanks!

PR #246, opened 2018-04-04
scanner: Fix detection of zero-length heredoc anchor.

octo:
The existing check did not take into account that a leading `-` is not considered part of the anchor.


mitchellh:
Thanks thanks!

Issue #237, opened 2018-03-02
HCL -> json equivalency example in readme is wrong?


ThisGuyCodes:
The readme says:
> ```hcl
> variable "ami" {
>     description = "the AMI to use"
> }
> ```
> This would be equivalent to the following json:
> ```js
> {
>   "variable": {
>       "ami": {
>           "description": "the AMI to use"
>         }
>     }
> }
> ```

However, based my usage of a [parser](https://github.com/kvz/json2hcl) (which admittedly might be out of date, though I built it with the latest version of this library) and the described functionality bellow:
> ... As an alternative, lists of objects can be created with repeated blocks, using this structure:
> ```hcl
> service {
>     key = "value"
> }
> 
> service {
>     key = "value"
> }
> ```

The actual equivalency appears to be:
```hcl
variable "ami" {
  description = "the AMI to use"
}
```
```js
{
  "variable": [
    {
      "ami": [
        {
          "description": "the AMI to use"
        }
      ]
    }
  ]
}
```

Are my assumptions correct? Is my parser wrong? At the very least the readme is unclear about how the `lists of objects can be created with repeated blocks` deals with only having one declaration.

gellweiler:
This is the workaround I currently use, because I need the extra [] to be gone:

```
flattenDataStructure(&objs)
data, e = json.Marshal(objs)

// Workaround for: https://github.com/hashicorp/hcl/issues/237, https://github.com/kvz/json2hcl/issues/3
func flattenDataStructure(data *map[string]interface{}) {
	for k, v := range *data {
		vListOfMap, ok := v.([]map[string]interface{})
		if ok {
			if len(vListOfMap) == 1 {
				(*data)[k] = vListOfMap[0]
			}
		}
	}

	for _, v := range *data {
		vMap, ok := v.(map[string]interface{})
		if ok {
			flattenDataStructure(&vMap)
		}
	}
}
```

It's not nice but at least for me it gets the job done.

PR #265, opened 2018-11-01
Incorrect equivalent JSON example

lyle-nel:
Equivalent JSON example is incorrect as per issue #237

This causes a lot of frustration for folks building these objects programmatically. The error returned by terraform, for example, is particularly opaque, which means it could take days to discover this undocumented caveat.


lyle-nel:
Hi Folks, any update on this?

PR #275, opened 2019-04-18
Fix unused keys

langmartin:
this fixes / implements the `hcl:",unusedKeys"` feature to track keys present in the hcl not used by the structs provided to guide decoding


decoder.go
```diff
@@ -404,16 +404,20 @@ func (d *decoder) decodeMap(name string, node ast.Node, result reflect.Value) er
...
 	resultType := result.Type()
-	resultElemType := resultType.Elem()
-	val := reflect.New(resultElemType)
+	val := reflect.Indirect(result)
+	if val.Kind() == reflect.Invalid || val == reflect.Zero(resultType) {
+		resultElemType := resultType.Elem()
+		val = reflect.New(resultElemType)
+		result.Set(val)
+	}
+
```
schmichael: I don't understand this change, but I'm also not an expert at reflection! :sweat_smile: 

The old code appears to:
1. Create a new instance of `result`'s type
2. Decodes into it
3. Sets `result` to that hydrated value

The new code appears to:
1. Indirects result to turn it into a non-pointer type
2. Checks if `val` is Invalid which... I don't think it can ever be? At least from the docs I don't see where an Invalid value could come from: https://golang.org/pkg/reflect/#Value
3. Checks if `val` is the zero value of `result`'s type which is only true if `result` is a nil point, so I think that could be shortened to `if result.IsNil()`?
4. Creates a zero value for `result`'s type which I think is what `Indirect(result)` does above already?
5. Sets result to the zero value.
6. Decodes as before (although the Indirect call is redundant AFAICT)
7. Does _not_ set the decoded value back to result... I guess perhaps reflect.Value has an internal pointer to the value so that this works properly?

So I'm not sure why the old method didn't work or why the new code works. Hopefully it's tested well enough to demonstrate it does and it's just my lack of reflection expertise at fault!

decoder.go
```diff
@@ -631,10 +635,19 @@ func (d *decoder) decodeStruct(name string, node ast.Node, result reflect.Value)
...
 	decodedFields := make([]string, 0, len(fields))
 	decodedFieldsVal := make([]reflect.Value, 0)
 	unusedKeysVal := make([]reflect.Value, 0)
+
+	// fill unusedNodeKeys with keys from the AST
+	// a slice because we have to do equals case fold to match Filter
+	unusedNodeKeys := make([]string, 0)
+	for _, item := range list.Items {
+		for _, k := range item.Keys {
+			unusedNodeKeys = append(unusedNodeKeys, k.Token.Value().(string))
```
schmichael: Is it safe to type assert `k.Token.Value()` to a `string`? The [docs](https://godoc.org/github.com/hashicorp/hcl/hcl/ast#ObjectKey) say it can be a string _or an identifier_. Is there a test exercising that this type assertion safe if the key is an identifier?

decoder_test.go
```diff
@@ -786,6 +786,40 @@ func TestDecode_structureMapInvalid(t *testing.T) {
...
+		Found: []string{"A", "B"},
+		Extra: []string{"extra1", "extra2"},
+	}
+
+	var p hclVariable
+	ast, _ := Parse(testReadFile(t, "structure_map_extra_keys.hcl"))
+	DecodeObject(&p, ast)
+	if !(p.A == q.A && p.B == q.B &&
+		reflect.DeepEqual(p.Found, q.Found) &&
+		reflect.DeepEqual(p.Extra, q.Extra)) {
```
schmichael: Why not just do `if !reflect.DeepEqual(p, q) {}`?

decoder.go
```diff
@@ -404,16 +404,20 @@ func (d *decoder) decodeMap(name string, node ast.Node, result reflect.Value) er
 }
 
 func (d *decoder) decodePtr(name string, node ast.Node, result reflect.Value) error {
-	// Create an element of the concrete (non pointer) type and decode
-	// into that. Then set the value of the pointer to this type.
+	// lookup the concrete (non pointer) type and decode into that
+	// if the pointer is nil, create an element of the type and set the pointer to that
```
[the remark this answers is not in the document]
langmartin: decodePtr overall exists to run decode on the reflect.Value wrapped around a concrete instance of the type pointed to. It's just reflect.Indirect, except that it may have to create the type. I'll take another swing at the comment.

decoder.go
langmartin: This change allows us to set default values. Before, when there was pointer in the config struct it would always create a new e.g. ClientConfig, even if we'd passed in one that had values. This code should reflect.Indirect unless the pointer is nil.

After I made this change I found https://github.com/hashicorp/hcl/pull/39 which is a much cleaner version of this same change, should I just incorporate that instead?

2. Experimentally, you get Kind Invalid when you indirect a nil pointer
3. I did mean IsNil, and that would clean all this up a bunch
4. I understood that I would get a zero value from Indirect, but I think there's a difference between primitive types and structs there. For a struct, you seem to get Kind Invalid
5. yeah, result holds the pointer, so we need to point it at the newly minted instance of the struct
6. I think it is redundant
7. correct, result is already a pointer to the passed in instance of the struct we're decoding into

The old method would just create new structs at every pointer whether there was already something there or not. The pr I referenced above did add a test that should go with this change, this is also exercised by the nomad use of the library.

decoder.go
langmartin: That's a great question, I was following the existing code in this method. Looking at the token code it's not always safe but I should be able to check before casting.

langmartin: for Keys, identifiers and strings are both strings, and we do want to call value because it may manipulate the string value

langmartin:
> Am I understanding correctly that `,unusedKeys` just didn't do anything at all before this PR, even though we had code to find it in the target struct? 🤔

Yes! I'm not sure if it was functional in an earlier version.

langmartin:
This includes @ianremmler 's fix from pr#39, reworked to match code changes since it was written.

PR #351, opened 2020-03-14
Make Format() collapse multiple trailing newlines into a single newline

acrewdson:
A long time ago I opened [this issue](https://github.com/hashicorp/hcl/issues/285), but it wasn't until today that I took a look at the hcl code for the first time. I was pleasantly surprised to discover that `Format()` appends a newline [here](https://github.com/hashicorp/hcl/blob/master/hcl/printer/printer.go#L64) after the actual pretty-printing has been done.

This made me think that while what @apparentlymart describes [here](https://github.com/hashicorp/hcl/issues/285#issuecomment-530932978) is definitely the better and more correct way to go, if `Format()` is already making this (slightly dirty) adjustment before it returns the result, maybe we could solve the multiple-newlines-at-EOF problem by just adjusting that adjustment. Maybe it's just dumb enough to work?


acrewdson:
realized this would need to target hcl2, so I'll close it

PR #352, opened 2020-03-16
Feature: ordered attribute names

x1unix:
Hello, HCL contributors :)

I'm working on open-source project [Gilbert](github.com/go-gilbert/gilbert/tree/gilbert2) - `make` replacement for Go project.

My project uses **hcl** to keep configuration about defined tasks and variables.
As my use-cases are much more complicated that *read-and-unmarshal file*, I process file manually with `hclsyntax.ParseConfig` and walk over document body.

Project's task file allows to declare tasks and variables (like in `makefile`) and one variable
can contain reference to another variable.

```hcl
foo = "foo"
bar = "${bar}"
```

Currently, `*hclsyntax.Body` stores all attributes in maps. It's okay for general purposes but introduces issues when attribute order matters because Go doesn't guarantee map order.

In that specific case I need to know in which order each attribute was declared to interpolate 
attribute value correctly (because attribute B may depend on value of attribute A).

```
[Eval attr A] -> [Add to context] -> [Eval Attr B]
```

To solve this issue, `hclsyntax.Body.AttributeNames()` was introduced.
It will return attribute names in in order which they were declared.

Currently I'm unable to find any valid workaround to get attributes declaration order, so I made this PR solve my problem (and I hope this will be useful for another people).

**Links for problem context**

* Example file mentioned in this PR: https://github.com/go-gilbert/gilbert/blob/gilbert2/v2/gilbert.hcl
* File processing code sample: https://github.com/go-gilbert/gilbert/blob/gilbert2/v2/manifest/processor.go#L133

### DIFF
--- a/hclsyntax/parser.go
+++ b/hclsyntax/parser.go
@@ -24,6 +24,7 @@ type parser struct {
 func (p *parser) ParseBody(end TokenType) (*Body, hcl.Diagnostics) {
 	attrs := Attributes{}
 	blocks := Blocks{}
+	var attrNames []string
 	var diags hcl.Diagnostics
 
 	startRange := p.PrevRange()
@@ -61,6 +62,7 @@ Token:
 					})
 				} else {
 					attrs[titem.Name] = titem
+					attrNames = append(attrNames, titem.Name)
 				}
 			default:
 				// This should never happen for valid input, but may if a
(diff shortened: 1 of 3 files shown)


x1unix:
@apparentlymart can you take a look at this PR please?

PR #359, opened 2020-03-26
Update link to godoc

glerb:


ext/userfunc/README.md
```diff
@@ -25,4 +25,4 @@ inclusion in a `hcl.EvalContext`. It also returns a new `cty.Body` that
 contains the remainder of the content from the given body, allowing for
 further processing of remaining content.
 
-For more information, see [the godoc reference](http://godoc.org/github.com/hashicorp/hcl/v2/ext/userfunc).
+For more information, see [the godoc reference](https://pkg.go.dev/github.com/hashicorp/hcl/v2/ext/userfunc?tab=doc).
```
DavidGamba: go.dev only supports released versions. It depends on the release cycle but sometimes it is nice to just be able to get a link to the godoc from master instead of waiting for the next release to see the doc fixes.

glerb: The existing link gives a "not found".

PR #361, opened 2020-04-17
Add windows tests to CircleCI

mdeggies:
This PR adds another job to run the go tests on a Windows executor, which will let us get rid of the appveyor dependency. To run tests with different versions of go, add the version(s) to the array's on lines 170 and 175. 

The linux and win tests run at the same time, but the Windows job does take a bit longer to finish due to the nature of the executor + some required setup steps. The workflow takes around ~1:23 to complete. The longest step (w/out caching) is installing go, which we have to do because the Windows executor only has go 1.12 installed. This job has gomod + coverage persistence + golang caching enabled, so it's a bit quicker. 

Besides removing the appveyor file, I also removed the travis config file in this PR, as the circle config seems to cover all tests for branch pushes/pr's/merges. 

Screen Shot 2020-04-16 at 10 37 07 PM

Let me know if you see anything wonky, and thanks for your review!


.circleci/config.yml
```diff
@@ -1,25 +1,21 @@
...
+    WIN_TEST_RESULTS: &WIN_TEST_RESULTS c:\Users\circleci\AppData\Local\Temp\test-results
+
+commands:
+  git-verify:
     steps:
-      - checkout
-      - run: go mod verify
-      - run: make fmt
-      - run:
-          name: verify no code was generated
```
alisdair, line 22: This is where I wrote a comment and GitHub seems to have deleted it for some reason, which is pretty rude. An attempt to reconstruct:

Looking at the test output, we're not running this command in the Linux tests. And in the Windows tests, we don't run `go mod verify` or `make fmt` before it, so it's always going to pass. It'd be great to have this check present and working on at least one of the platforms.

mdeggies: Ahh good catch! I forgot to add that job to the workflows. I'll have the check job run before the tests on windows & linux.

mdeggies: Fixed & squashed my commits- let me know how it looks! Can cherrypick once this is merged or open another PR (unless that's overkill). 

Screen Shot 2020-04-21 at 6 13 19 PM

PR #370, opened 2020-04-29
Add tokens.Equal method for use in tests or for usage detection

bendrucker:
This PR adds an `Equal` method that can check if one `tokens` slice is equal to another. I found myself [writing](https://github.com/bendrucker/terraform-cloud-migrate/blob/1a10af5696b831528c73c9dc79ba142fec858c88/variable.go#L172-L185) this method to detect whether a `terraform.workspace` expression was present in any attribute before using `RenameVariablePrefix` (in order to track the files where replacement was needed). 

It can be used in the tests here and might also be useful for tests that users write against their own code that uses hclwrite.

### DIFF
--- a/hclwrite/ast_body_test.go
+++ b/hclwrite/ast_body_test.go
@@ -2,7 +2,6 @@ package hclwrite
 
 import (
 	"fmt"
-	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


bendrucker:
I've realized that using this probably isn't the right approach for my project so I'm going to close and avoid adding something to the public API that perhaps doesn't need to be there.

PR #372, opened 2020-05-06
hclwrite: Generate multi-line objects and maps

alisdair:
The previous syntax for object and map values was a single line of key-value pairs. For example:

```hcl
object = { bar = 5, baz = true, foo = "foo" }
```

This is very compact, but in practice for many HCL values, less readable and less common than a multi-line object format. This commit changes the generated output from hclwrite to one line per attribute.

Examples of the new format:

```hcl
// Empty object/map is a single line
a = {}

// Single-value object/map has the attribute on a separate line
b = {
  bar = 5
}

// Multi-value object/map has one line per attribute
c = {
  bar = 5
  baz = true
}
```

### DIFF
--- a/hclwrite/examples_test.go
+++ b/hclwrite/examples_test.go
@@ -48,9 +48,13 @@ func Example_generateFromScratch() {
 	// Output:
 	// string = "foo"
 	//
-	// object = { bar = 5, baz = true, foo = "foo" }
-	// bool   = false
-	// path   = env.PATH
+	// object = {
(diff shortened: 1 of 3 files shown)


alisdair:
Thanks, @mildwonkey! Based on your review and a conversation with @apparentlymart yesterday, I feel fairly confident that this is a reasonable thing to do. It's also blocking a Terraform PR, so I'll merge & release, and undo it if it turns out to have been a bad idea.

PR #383, opened 2020-05-28
hclsyntax: Introduce token-based parse methods

radeksimko:
This change introduces new methods to allow two-phased approach where tokenization is done prior to parsing.

This is specifically useful in the context of a language server, so it can pass around a single interpretation of HCL blocks (tokens) instead of passing around both block and its tokens.

Related: https://github.com/hashicorp/terraform-ls/pull/125

### New methods

```go
func ParseBodyFromTokens(tokens Tokens, end TokenType) (*Body, hcl.Diagnostics)
func ParseBodyItemFromTokens(tokens Tokens) (Node, hcl.Diagnostics)
func ParseBlockFromTokens(tokens Tokens) (*Block, hcl.Diagnostics)
func ParseAttributeFromTokens(tokens Tokens) (*Attribute, hcl.Diagnostics)
```

### DIFF
--- a/hclsyntax/peeker.go
+++ b/hclsyntax/peeker.go
@@ -112,7 +112,11 @@ func (p *peeker) nextToken() (Token, int) {
 	// if we fall out here then we'll return the EOF token, and leave
 	// our index pointed off the end of the array so we'll keep
 	// returning EOF in future too.
-	return p.Tokens[len(p.Tokens)-1], len(p.Tokens)
+	return p.lastToken(), len(p.Tokens)
+}
+
+func (p *peeker) lastToken() Token {
+	return p.Tokens[len(p.Tokens)-1]
 }
 
 func (p *peeker) includingNewlines() bool {
--- a/hclsyntax/public.go
+++ b/hclsyntax/public.go
@@ -1,6 +1,8 @@
 package hclsyntax
 
 import (
+	"fmt"
+
 	"github.com/hashicorp/hcl/v2"
 )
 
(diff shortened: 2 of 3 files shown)


radeksimko:
After testing it a bit more I think it would be sensible to add some checks to `ParseBodyItemFromTokens`, I was specifically tripped up by not knowing the token sequence:
 - must start with `TokenIdent`, e.g. newline must be stripped if there is one
 - must end with either `TokenEOF` or `TokenNewline`, i.e. one has to be inserted if sequence ends with `TokenCBrace`

radeksimko:
Regarding `ParseBodyItemFromToken` and the two other similar methods:

TL;DR language server currently uses these methods, mostly as a form of optimization.

I will try to share some more context below.

The main reason behind tokenization done in a separate step is that language server needs to know where _exactly_ in a parsed token `hcl.Pos` is, so it can provide correct completion data (e.g. complete just `ource` when user requested completion in `res<HERE>`, rather than whole `resource` which would result in `resresource`).

This can be implemented a few ways

 - maintain a single long array of tokens for all the configuration and perform lookup there
 - maintain tokens for each block and perform much more scoped lookup in much shorter array of tokens representing just that single block (this is what the language server does today)

With that said I understand your concerns and I admit this may be unnecessary pre-optimization which doesn't come cheap from API perspective. Perhaps a single array is just fine.

--- 

I'm currently working on a proposal for a new parser for the language server, which also involves looking into this problem more deeply. I have not figured out yet how to provide accurate completion data within `hcl.Expression` where the expression is made of more than a single token. This may affect the API a bit too, so I'll chew on that a bit more and come back to this PR.

radeksimko:
I'm going to close this as I found a different (more elegant?) way of resolving the same problem. The new API which was just published in the form of `hashicorp/hcl-lang` accepts `*hcl.File` which in itself holds raw bytes of each file and that allows me to re-tokenize these bytes on-demand and do so only when really necessary (e.g. when looking up for completion candidates) and only for the relevant file.

PR #385, opened 2020-06-01
hclsyntax: ConditionalExpr unification fallback

alisdair:
If we are unable to unify the two sides of a conditional expression, this should not result in an error. Instead we should return the unconverted value, and if this causes an error higher up in the evaluation, it can be handled there.

### Motivation

The motivation for this is to allow disjoint conditional expression values to be evaluated, where the context of the expression permits this. For example, the `merge` function can combine any object types, and we should be able to use conditional expressions as arguments to this function. [See this downstream Terraform issue for more real world examples](https://github.com/hashicorp/terraform/issues/22405).

This commit also updates the go-cty dependency in order to bring in the `merge` function for a test case of the new behaviour.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/sergi/go-diff v1.0.0
 	github.com/spf13/pflag v1.0.2
 	github.com/stretchr/testify v1.2.2 // indirect
-	github.com/zclconf/go-cty v1.2.0
+	github.com/zclconf/go-cty v1.4.2
 	golang.org/x/crypto v0.0.0-20190426145343-a29dc8fdc734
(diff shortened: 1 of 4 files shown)


alisdair:
This isn't quite right. Will update with more details later.

alisdair:
The reason this isn't right is because we explicitly do not want the type of an expression to change over time, as this introduces subtle bugs which are hard to diagnose. This commit would have made the conditional expression return type variable, which could cause distant errors as the return value is used in other expressions.

We must ensure that the type of expressions is predictable, even when unknown values are involved, which means that this simplistic solution to the problem is not an acceptable one. I'm still working on improved diagnostics for the Terraform bug which triggered this work.

PR #386, opened 2020-06-03
hclsyntax: address multiple issues with sequences (...)

mildwonkey:
Update: I ended up tackling two adjacent bugs at once. 

-----

Previously functions such as `concat()` would result in a panic if there
was a null element and a sequence, as in the included test. This PR adds
a check if the error index is outside of the range of arguments and
crafts an error that references the entire function instead of the null
argument.

Tested locally to confirm that this fixes a reported terraform panic:
Screen Shot 2020-06-03 at 9 13 08 AM

### DIFF
--- a/hclsyntax/expression.go
+++ b/hclsyntax/expression.go
@@ -260,6 +260,8 @@ func (e *FunctionCallExpr) Value(ctx *hcl.EvalContext) (cty.Value, hcl.Diagnosti
 		}
 
 		switch {
+		case expandVal.Type().Equals(cty.DynamicPseudoType):
+			return cty.DynamicVal, diags
 		case expandVal.Type().IsTupleType() || expandVal.Type().IsListType() || expandVal.Type().IsSetType():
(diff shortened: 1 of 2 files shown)


mildwonkey:
https://github.com/hashicorp/terraform/issues/24910 is the reference for the second commit.

hclsyntax/expression.go
```diff
@@ -260,6 +260,8 @@ func (e *FunctionCallExpr) Value(ctx *hcl.EvalContext) (cty.Value, hcl.Diagnosti
 		}
 
 		switch {
+		case expandVal.Type().Equals(cty.DynamicPseudoType):
```
jbardin, line 263: If passing `null` as the first argument isn't valid, we should have the same null check here as below. Otherwise we have a known value turning into an unknown value.

mildwonkey: thanks, fixed!

PR #390, opened 2020-06-16
hclwrite: Add block label quoting to format

alisdair:
Blocks consist of an identifier, followed by zero or more labels, and a body. Each label can be either an identifier or a quoted string literal.

This commit updates `hclwrite`'s format logic to replace any identifier labels with quoted strings. That is, for this input:

```hcl
resource null_resource test { }
```

`hclwrite` will prduce the following output:

```hcl
resource "null_resource" "test" { }
```

### DIFF
--- a/hclwrite/format.go
+++ b/hclwrite/format.go
@@ -15,12 +15,13 @@ var nilToken = &Token{
 }
 
 // format rewrites tokens within the given sequence, in-place, to adjust the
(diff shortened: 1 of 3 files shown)


alisdair:
Ah, I hadn't even considered that approach! Yes, that seems much less likely to be a disaster. Closing this and I'll try to help out with #340 when I can.

PR #393, opened 2020-07-15
Assignment of capsule type values from expressions in gohcl

dbolotin:
To allow assignment of fields in structs like:

```go
type Endpoint struct {
	// ...
	ForwardTo chan string `hcl:"to"`
}
```

with code like

```hcl
endpoint "example" {
  to = chan_variable
}
```


dbolotin:
Unfortunately, this approach turned out to be far from optimal.

PR #404, opened 2020-09-22
hclsyntax: Pass marks through template expressions

alisdair:
If a template expression interpolates values which have marks, we should apply all of those marks to the output value. This allows template expressions to function like native cty functions with respect to marks.

### DIFF
--- a/hclsyntax/expression_template_test.go
+++ b/hclsyntax/expression_template_test.go
@@ -283,6 +283,27 @@ trim`,
 			cty.UnknownVal(cty.String),
 			1, // Unexpected endfor directive
 		},
+		{ // marks from uninterpolated values are ignored
+			`hello%{ if false } ${target}%{ endif }`,
+			&hcl.EvalContext{
+				Variables: map[string]cty.Value{
+					"target": cty.StringVal("world").WithMarks(cty.NewValueMarks("sensitive")),
+				},
+			},
+			cty.StringVal("hello"),
+			0,
+		},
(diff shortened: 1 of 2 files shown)


pselle (approved):
🎉

hclsyntax/expression_template_test.go
```diff
@@ -283,6 +283,27 @@ trim`,
 			cty.UnknownVal(cty.String),
 			1, // Unexpected endfor directive
 		},
+		{ // marks from uninterpolated values are ignored
+			`hello%{ if false } ${target}%{ endif }`,
+			&hcl.EvalContext{
+				Variables: map[string]cty.Value{
+					"target": cty.StringVal("world").WithMarks(cty.NewValueMarks("sensitive")),
```
azr, line 290: Howdy ! I was looking at this and nice !!! We are probably going to use that for Packer too :D.

Question: can an object or a list be sensitive ? I couldn't tell from the test here ( also maybe ); ex:

```hcl
variable "foo" {
  default = { 
    key = "secret"
    value = "secret"
  }
  sensitive = true
}
```

alisdair: I think that's our hope, although it's very much work in progress. Applying marks to complex values currently doesn't work reliably, and this specific case (assuming your interpolation is like `"key=${var.foo.key}"`) will currently panic.

I'll be opening a follow-up PR for that issue, and should be able to add an interpolation test case then. Thanks!

PR #412, opened 2020-10-22
"body" struct tag to capture entire body for a block decode

mitchellh:
Structs can specify the "body" struct tag to capture the entire block
body. This does NOT act as "remain" where leftover fields are
non-erroneous. If you want leftover fields to be allowed, a "remain"
field must ALSO be present.

This is used to capture the full block body that was decoded for the
block. This is useful if you want to ever redecode something differently
(maybe with a different EvalContext) or partially decode something but
redecode the entire value.


apparentlymart (approved):
This looks good to me! I'm glad it was relatively easy to implement. :+1:

PR #415, opened 2020-10-23
Decode blocks into map with custom struct types

notnoop:
HCLv2 gohcl supports parsing labeled blocks into slices (e.g. `[]customStruct`).  This adds support for decoding labeled blocks into map types (e.g. `map[string]customStruct`), with the additional constraint that labels are unique.

We special case single labeled blocks, as the behavior for multi labeled blocks is not well scoped.

This i s motivated by some map typed fields in Nomad structure (like [`TaskGroup.Volumes`](https://github.com/hashicorp/nomad/blob/9e9dac4a2c94deffaf3857d8080b947f112c9b11/api/tasks.go#L420), and [`ConsulGatewayProxy. EnvoyGatewayBindAddresses`](https://github.com/hashicorp/nomad/blob/9e9dac4a2c94deffaf3857d8080b947f112c9b11/api/services.go#L347). While I suspect this is of wider benefit, I understand if the use case is deemed to narrow to include into the base gohcl library.

Concretely, assuming a type `Policies  map[string]customStruct 'hcl:"policy,block"'`, the following hcl declaration

```hcl
policy "foo" {
  val = "a"
}
policy "bar" {
  val = "b"
}
```

is decoded as

```go
map[string}*customStruct{
  "foo": &customStruct{Val: "a"},
  "bar": &customStruct{Val: "b"},
}
```

### DIFF
--- a/gohcl/decode.go
+++ b/gohcl/decode.go
@@ -136,6 +136,7 @@ func decodeBodyToStruct(body hcl.Body, ctx *hcl.EvalContext, val reflect.Value) 
 		ty := field.Type
 		isSlice := false
 		isPtr := false
+		isMap := false
 		if ty.Kind() == reflect.Slice {
 			isSlice = true
 			ty = ty.Elem()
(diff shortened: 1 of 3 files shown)


thejohnbackes:
Is there any movement on this PR? I have a need for supporting labeled blocks, and the workaround is... ugly.

PR #427, opened 2020-12-03
Unmark conditional values before checking their truthiness

pselle:
In conditional expressions that involve a marked value, unmark the value before inspecting its truthiness. This avoids a mark panic in cty when checking the value. Related to https://github.com/hashicorp/terraform/issues/27107


mildwonkey (approved):
`truthiness` 😂

PR #435, opened 2020-12-16
Add win tests to CircleCI config and remove Appveyor

mdeggies:
- Adds the windows tests to CircleCI to remove the dependency on Appveyor (which is being deprecated)
- CI config is pretty similar to what's in master now (e.g. we are using gotestsum for windows tests as well, which has a more verbose output to help in the case of test failures)


alisdair (approved):
Thanks!

PR #436, opened 2020-12-18
hclsyntax: Fix panic for marked collection splat

alisdair:
Fixes [this downstream Terraform bug](https://github.com/hashicorp/terraform/issues/27329). I couldn't find any other instances of calling `LengthInt` on a possibly-marked collection.

### DIFF
--- a/hclsyntax/expression_test.go
+++ b/hclsyntax/expression_test.go
@@ -1172,6 +1172,26 @@ upper(
 			}).Mark("sensitive"),
 			0,
 		},
+		{ // splat with collection with sensitive elements
+			`maps.*.x`,
+			&hcl.EvalContext{
+				Variables: map[string]cty.Value{
+					"maps": cty.ListVal([]cty.Value{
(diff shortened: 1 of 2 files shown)


hclsyntax/expression_test.go
```diff
@@ -1156,7 +1156,22 @@ upper(
 			}),
 			1,
 		},
-
+		{ // splat with sensitive collection
+			`maps.*.enabled`,
+			&hcl.EvalContext{
+				Variables: map[string]cty.Value{
+					"maps": cty.ListVal([]cty.Value{
+						cty.MapVal(map[string]cty.Value{"enabled": cty.True}),
```
pselle, line 1164: I think splat in the scenario where a member is marked but the collection itself isn't impacted by this bug/change, but this could be a good opportunity to add that test coverage if not already present.

pselle (approved):
👍 with a possible test expansion, if wanted/warranted

PR #438, opened 2020-12-18
hclsyntax: Fix for expression marked conditional

alisdair:
These test cases both panic without the associated changes, because `.False()` on a marked value is invalid.

If the conditional expression in a for expression evaluates to a marked value, we need to ignore the marks before testing it. Since this result will not interpolate into the result, dropping the mark here seems reasonable for now.

### DIFF
--- a/hclsyntax/expression.go
+++ b/hclsyntax/expression.go
@@ -1143,7 +1143,17 @@ func (e *ForExpr) Value(ctx *hcl.EvalContext) (cty.Value, hcl.Diagnostics) {
 					continue
 				}
 
-				if include.False() {
+				// Extract and merge marks from the include expression into the
+				// main set of marks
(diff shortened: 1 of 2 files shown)


apparentlymart (approved):
Given that the "marks" feature in cty is intended as a general thing rather than specifically for Terraform's sensitive feature I might argue that it's better to be conservative here and preserve the union of marks from all three operands in the result, but I can also see your argument to do the opposite here and I don't have a _strong_ opposition to it.

One thing I would say though is that adding new marks in a place where there weren't before is likely to be a more disruptive change for a consuming application than the opposite action of making something no longer be marked where it was before, so if we're unsure I'd suggest erring on the side of preserving marks in order to better keep options open for later.

alisdair:
Thanks, that makes sense! I've updated the commit to preserve the marks from the conditional in the final value.

PR #440, opened 2021-01-04
Mark objects with keys that are sensitive

pselle:
This adjusts prior behavior that would error to now allow keys that are marked; those marks are removed in evaluation in order to give a valid string for the map key, but the resulting object or tuple is marked as a whole with whatever mark the key contained, ensuring marks are maintained.

Follow-on work to #434


alisdair (approved):

apparentlymart (approved):
This looks good to me! Thanks for doing this follow-up.

PR #444, opened 2021-02-08
add an ExprMap method to the mock expression

jbardin:
Add an ExprMethod so that the mock expressions can be handled by the
hcl.ExprMap function.

Added basic tests for both ExprList and ExprMap

### DIFF
--- a/hcltest/mock_test.go
+++ b/hcltest/mock_test.go
@@ -1,6 +1,7 @@
 package hcltest
 
 import (
+	"strings"
 	"testing"
 
 	"reflect"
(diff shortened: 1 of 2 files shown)


hcltest/mock_test.go
```diff
@@ -266,3 +269,138 @@ func TestMockBodyPartialContent(t *testing.T) {
...
+					t.Errorf("expected error %q, got %q", tc.Diags, diags)
+				}
+				if !diags.HasErrors() {
+					t.Errorf("expected diagnostic message %q", tc.Diags)
+				}
+			} else if diags.HasErrors() {
+				t.Error(diags)
+			}
+
+			// map keys may be arbitrarily ordered
```
alisdair, line 392: I could be wrong and/or misunderstanding this comment, but I think cty map & object iteration order is intentionally consistent ([see here](https://github.com/zclconf/go-cty/blob/251ac0b66ee75102f9e2dd88a1ae1743d3dd421f/cty/element_iterator.go#L53-L60)). If so I think it would be safe to depend on that ordering if you wanted to simplify this test slightly.

alisdair (approved):
Looks good to me! I have a small note on the sorting in the test, but I think the existing code is perfectly fine so feel free to ignore.

PR #451, opened 2021-02-26
avoid panics with marked Index key values

jbardin:
Index keys must be unmarked to convert them to a primitive type.

Add tests for overall `Index` test coverage.


alisdair (approved):
Looks good to me! Thanks for documenting so many edge cases with tests.

PR #452, opened 2021-03-07
Update spec.md

ahgittin:
Fix inconsistent vars


radeksimko (approved):
Good catch, thanks for the PR!

PR #453, opened 2021-03-10
Fix wrong var name in HCL guide

antoineco:
The example below taken from the guide

> ```hcl
> variable "network_numbers" {
>   type = list(number)
> }
> ```
>
> In this example, the variable "network_numbers" block makes `var.base_network_addr` available to expressions,

should instead read:

```diff
In this example, the `variable "network_numbers"` block makes
- `var.base_network_addr`
+ `var.network_numbers`
available to expressions,
```

### DIFF
--- a/guide/go_patterns.rst
+++ b/guide/go_patterns.rst
@@ -47,8 +47,8 @@ definitions of values that are available for use in expressions elsewhere:
      value = cloud_subnet.example[*].id
    }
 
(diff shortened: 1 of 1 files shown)


antoineco:
cc @jbardin @apparentlymart (I think you are the most active maintainers of this repo?)

radeksimko (approved):
Good catch, thank you for the PR!

PR #455, opened 2021-03-26
userfunc: allow constructing a single function

tonistiigi:
Refactors previous internal functionality of decoding
a single function so that functions can be optionally
decoded individually.

This is useful then functions need to depend on each
other and static analysis is performed to understand
what variables/functions they depend on.

No functional changes to existing `DecodeUserFunctions`.

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

### DIFF
--- a/ext/userfunc/decode.go
+++ b/ext/userfunc/decode.go
@@ -2,7 +2,6 @@ package userfunc
 
 import (
 	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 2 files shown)


crazy-max (commented):
@tonistiigi Tested on my side and LGTM. Maybe we should wait for hashicorp/hcl#455 before merge?

tonistiigi:
@radeksimko @jbardin

PR #463, opened 2021-04-22
ext: TryFunc tests to document handling of marked values

mildwonkey:
This PR adds some test coverage to illustrate handling marked inputs in `TryFunc`.  `TryFunc` preserve marks on return values, including preserving marks on collections if the result is an element of said collection.


mildwonkey:
I let releng know that this repository needs a new deploy key for circleci

alisdair (approved):

PR #466, opened 2021-04-30
Backport unused key tracking from main to the Vault fork branch

sgmiller:
... and add file position tracking.


HridoyRoy (approved):
lgtm!

go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/hashicorp/hcl
 
+go 1.15
```
HridoyRoy, line 3: Quick question, I noticed that go.mod's go version was upgraded to go 1.15, which I believe means the minimum version required to use the hcl library is 1.15 right? Are all the plugins using this hcl library on version 1.15?

PR #467, opened 2021-05-04
Fix parseQuotedStringLiteral zero end range

alisdair:
If a block label had an invalid quoted string literal, the parser would previously create label ranges with uninitialized `End` values. For Terraform, this would eventually percolate into invalid diagnostic output due to the assumption that the end of a range is always greater than or equal to the start: see https://github.com/hashicorp/terraform/issues/28587 and https://github.com/hashicorp/terraform/pull/28598

This commit addresses this by using the terminating token's position as the end of the range, even if it's not a closing quote.

### DIFF
--- a/hclsyntax/parser.go
+++ b/hclsyntax/parser.go
@@ -1661,7 +1661,7 @@ func (p *parser) parseQuotedStringLiteral() (string, hcl.Range, hcl.Diagnostics)
 
 	var diags hcl.Diagnostics
 	ret := &bytes.Buffer{}
-	var cQuote Token
+	var endRange hcl.Range
 
 Token:
 	for {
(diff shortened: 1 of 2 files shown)


mildwonkey (commented):
I am completely in favor of this, but I wonder if it should be considered a breaking change for other consumers who might be checking for this specifically. What do you think?  🤔

alisdair:
Hmm, it's fuzzy. I think I'd lean towards making this a patch release change. The prior behaviour feels both unintentional and broken to me, which makes it a bug, which we don't have to consider a compatibility surface.

I looked at the Packer and Nomad uses of this code path, and neither of them does this kind of inspection of the diag outputs. Packer uses HCL's own diagnostic text writer, and Nomad uses the `Error()` method to print them.

mildwonkey (approved):
That sounds good to me, thanks!

PR #485, opened 2021-08-13
hclsyntax: Fix typo

lemon-li:


radeksimko (approved):
Good catch, thanks for the PR!

PR #487, opened 2021-09-22
Align `hcl.Block` & `hclsyntax.Block` `DefRange`

radeksimko:
This patch fixes a minor discrepancy between `hclsyntax` and `hcl` in how DefRange is computed for a block.

DefRange of the "higher-level" `hcl.Block` (typically used by most applications) ends with the last label.

DefRange of the "lower-level" `hclsyntax.Block` ended with the opening brace `{` (i.e. at least 2 characters further).

Here we align the low-level API with the high-level one, meaning that most consumers should not be affected by this change, partly because they use `hcl.Block` (which isn't changing) or because they use DefRange to inform some non-critical decisions such as enriching diagnostics.

--- 
I ran into this discrepancy when working on adding support to JSON syntax to the language server in https://github.com/hashicorp/hcl-lang/pull/88

### DIFF
--- a/hclsyntax/navigation_test.go
+++ b/hclsyntax/navigation_test.go
@@ -112,12 +112,12 @@ data "another" "baz" {
 	}{
 		{0, hcl.Range{}},
 		{2, hcl.Range{}},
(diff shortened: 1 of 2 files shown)


apparentlymart (approved):
Makes sense to me!

PR #488, opened 2021-09-22
hclsyntax: Clarify Body.EndRange comment

radeksimko:
As discussed via Slack this field is no longer (and probably never was) used for reporting missing items.

Instead `Body` has a dedicated method for this:

https://github.com/hashicorp/hcl/blob/2ebbb5d2dcc67881774d2b81d1497a55452e890f/hclsyntax/structure.go#L281-L287

### DIFF
--- a/hclsyntax/structure.go
+++ b/hclsyntax/structure.go
@@ -40,7 +40,7 @@ type Body struct {
 	hiddenBlocks map[string]struct{}
 
 	SrcRange hcl.Range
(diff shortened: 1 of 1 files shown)


apparentlymart (approved):
We did have at least a few Terraform releases where this comment was true, but we switched how `hclsyntax.Body` implements `MissingItemRange` after realizing that the "def range" gives better context for a problem when we highlight it as part of a diagnostic message and so indeed this comment is stale now.

PR #493, opened 2021-11-18
splat expression upgrading an unknown value should be unknown

jbardin:
When a splat expression is used to upgrade an unknown value to a list with a single value, the result must be entirely unknown. 

Using a splat expression to expand a single non-list value will result in a list with a single value, unless that value is null, in which case it will result in an empty list. This means that if `ukstr` is an unknown string value, then `unkstr[*]` could either result in a list of 1 string value, or an empty list, hence the result of `unkstr[*]` must be entirely unknown.

### DIFF
--- a/hclsyntax/expression.go
+++ b/hclsyntax/expression.go
@@ -1432,9 +1432,22 @@ func (e *SplatExpr) Value(ctx *hcl.EvalContext) (cty.Value, hcl.Diagnostics) {
 		return cty.DynamicVal, diags
 	}
 
+	upgradedUnknown := false
 	if autoUpgrade {
+		// If we're upgrading an unknown value to a tuple/list, the result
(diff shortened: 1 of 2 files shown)


apparentlymart (approved):
Thanks for looking into this, @jbardin!

It does feel like we've been gradually stepping towards a broad rule "if the given value is unknown then the result is just immediately `cty.DynamicVal`" by gradually finding weird ways that our more-specific type inference rules weren't correct, but what you did here does seem correct to me and I'd rather keep the precise type information if we can, so mainly this paragraph is just me telling myself that if we find _yet another_ of these weird exceptions then next time the answer is probably just a blanket rule and accept that the behavior of this operator is complex enough that it's justified to discard type information and just figure things out at "runtime". (that is, once all of the values are known)

PR #495, opened 2021-12-01
unknown upgraded splat values may have no elements

jbardin:
Follow up for #493, which ensured the value of an upgraded splat expression on an unknown value is unknown, but could result in the wrong type. When the previous PR was updated to allow specific types for expressions with attribute selectors after the splat, the resulting types were still unknown tuples even though the number of elements could not be determined. 

When upgrading an unknown splat value, the resulting collection may have 0 elements if the value ends up being `null`. Since the final result may be a tuple type with 0 or 1 elements, the unknown value must also be dynamic.


apparentlymart (approved):
The loss of the type information here is unfortunate for sure, but I think it makes pragmatic sense given the various different permutations this operator supports in order to match as closely as possible how Terraform v0.11 and earlier treated `.*` in resource references; it isn't really a thing that was designed with precise type information in mind, after all.

PR #497, opened 2021-12-02
hclsyntax: test for invalid whitespace heredoc

kmoe:
This PR adds two unit tests demonstrating invalid heredoc syntax, intended as a discussion point.

## Context

The following Terraform variable definition yields an error (https://github.com/hashicorp/terraform/issues/25410):
```hcl
policy = <<EOF 
{
  "hello": "hi"
}
EOF
```
Note the single whitespace character following "<<EOF" on the first line. Adding a matching whitespace character after the final "EOF" does not make it valid.

This is correct behaviour according to the HCL [spec](https://github.com/hashicorp/hcl/commit/2ebbb5d2dcc67881774d2b81d1497a55452e890f):
```ebnf
TemplateExpr = quotedTemplate | heredocTemplate;
quotedTemplate = (as defined in prose above);
heredocTemplate = (
    ("<<" | "<<-") Identifier Newline
    (content as defined in prose above)
    Identifier Newline
);
```

Since `Identifier` cannot contain whitespace, the first line does not introduce a heredoc template expression. A heredoc cannot be introduced by a line ending in whitespace.

### Bash

Heredocs in Bash scripts behave the same way as HCL when encountering trailing whitespace. The following is not a valid heredoc:
```bash
cat << EOF 
myheredoc
EOF
```

Executing the above lines in an interactive Bash shell one-by-one, however, will result in a valid heredoc, since the shell strips trailing whitespace.

### Perl

Perl seems to allow trailing whitespace in the heredoc delimiter, so the following is valid:
```perl
print <<"END ";
myheredoc
END 
```

while the following is not, since there is no whitespace at the end of the terminating delimiter:
```perl
print <<"END ";
myheredoc
END
```

## Proposal

HCL's current behaviour is reasonable, and follows the ubiquitous Bash precedent, at the expense of some of Perl's flexibility (and eccentricity). Adding these tests will make it clearer to consumers. 

A better resolution to https://github.com/hashicorp/terraform/issues/25410 could involve a more specific error message, but I think the parser would need to do some sort of similarity calculation, which is a larger design question.

### DIFF
--- a/hclsyntax/scan_tokens_test.go
+++ b/hclsyntax/scan_tokens_test.go
@@ -1508,6 +1508,150 @@ EOT
 			},
 		},
 		{
+			`<<EOT
+  hello world
+EOT
+`,
+			[]Token{
+				{
+					Type:  TokenOHeredoc,
+					Bytes: []byte("<<EOT\n"),
+					Range: hcl.Range{
+						Start: hcl.Pos{Byte: 0, Line: 1, Column: 1},
+						End:   hcl.Pos{Byte: 6, Line: 2, Column: 1},
+					},
+				},
+				{
+					Type:  TokenStringLit,
+					Bytes: []byte("  hello world\n"),
+					Range: hcl.Range{
+						Start: hcl.Pos{Byte: 6, Line: 2, Column: 1},
+						End:   hcl.Pos{Byte: 20, Line: 3, Column: 1},
+					},
+				},
+				{
+					Type:  TokenCHeredoc,
+					Bytes: []byte("EOT"),
+					Range: hcl.Range{
+						Start: hcl.Pos{Byte: 20, Line: 3, Column: 1},
+						End:   hcl.Pos{Byte: 23, Line: 3, Column: 4},
+					},
+				},
+				{
+					Type:  TokenNewline,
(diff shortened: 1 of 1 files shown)


radeksimko:
Here is another example which may be worth testing for - leading whitespace in front of the closing heredoc token:
https://go.dev/play/p/OytUIvTYAsZ

```hcl
policy = <<EOF
{
  "hello": "hi"
}
  EOF
```

It appears that the whitespace is just being ignored by the lexer - which is likely by design - but best to codify that decision into a test.

kmoe:
Thanks! I've added the following test cases, to cover various leading whitespace cases:

1
```hcl
<<EOT
  hello world
EOT
```

2
```hcl
<<-EOT
  hello world
EOT
```

3
```hcl
<<-EOT
  hello world
 EOT
```

As can be seen in the unit tests, the lexer does not in fact ignore the whitespace - this is handled by the parser. 

While adding a test case to `parser_test.go` I discovered that the deep-equal comparisons in that test were not working properly, which I'll raise another PR to fix. Until then the only tests explicitly confirming that leading whitespace is ignored by the _parser_ are in the specsuite.

alisdair (approved):

PR #521, opened 2022-03-26
Allows assigning labels to custom types having string as underlying type

mariomac:
Given a body using a label from a custom type that has `string` as underlying type:

```go
type Custom string
type Body struct {
    Block struct {
        Name Custom `hcl:"name,label"`
    } `hcl:"block,block"`
}
```

The following HCL would fail due to a reflection error: 

```hcl
block "foo" {}
```

```
Error: reflect.Set: value of type string is not assignable to type gohcl.Custom
```

With the patch in this PR, this operation is now accepted and the `Name` field would get the `foo` label value.


mariomac:
Hi! I still think this PR is relevant for  many Go projects but I don't see any activity here for months. Shall I close it?

crw:
Hi @mariomac, I'll try to get an answer for you as to whether this can be reviewed or not. Thanks!

PR #526, opened 2022-04-13
gohcl: skip encoding additional nil fields

jedevc:
This patch skips encoding nil slices, maps and interfaces, in addition to the already skipped nil pointers.

The scenario I've run into this in is something like the following:

```go
type X struct {
	A *string  `hcl:"a"`
	B []string `hcl:"b"`
}

...

f := hclwrite.NewEmptyFile()
block := gohcl.EncodeAsBlock(X{}, "test")
f.Body().AppendBlock(block)
fmt.Println(string(f.Bytes()))
```

Without the patch I get:
```hcl
test {
  b = null
}
```
While with the patch, I get:
```
test {
}
```

Hopefully this small change is something worth incorporating!


jedevc:
Is there anything currently blocking this?

yuri-tceretian:
I think the change the change does a good thing. Any chance it can be merged?

PR #532, opened 2022-05-24
Detect and warn on duplicate map keys

alexjurkiewicz:
This PR is currently WIP. I want to get early feedback before continuing. Is this something the HCL maintainers would accept?

(I'm also stuck technically on how to implement the below points -- would love some advice 🙏.)

TODO:
- [ ] Add tests - current tests use JSON input, which doesn't allow duplicate map keys. Create tests using HCL input to validate behaviour.
- [ ] Add context to warning


alexjurkiewicz:
Related: #35

PR #549, opened 2022-08-23
typeexpr: Optional object attributes with defaults

alisdair:
This commit extends the type expression package to add two new features:

- In constraint mode, the `optional(...)` modifier can be used on object attributes to allow them to be omitted from input values to a type conversion process. Any such missing attributes will be replaced with a `null` value of the appropriate type upon conversion.
- In the new defaults mode, the `optional(...)` modifier takes a second argument, which accepts a default value of an appropriate type. These defaults are returned alongside the type constraint, and may be applied prior to type conversion through the new `Defaults.Apply()` method.

This change is upstreamed from Terraform, where optional object attributes have been available for some time. [The defaults functionality is new](https://github.com/hashicorp/terraform/pull/31154) and due to be released with Terraform 1.3.

### DIFF
--- a/ext/typeexpr/defaults.go
+++ b/ext/typeexpr/defaults.go
@@ -1,0 +1,157 @@
+package typeexpr
+
+import (
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Defaults represents a type tree which may contain default values for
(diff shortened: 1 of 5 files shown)


kmoe (approved):
LGTM!

PR #550, opened 2022-08-28
Update spec.md to avoid suggesting ! is a binary operator

ascopes:
The current spec appears to suggest that the bang is a binary operator, which would enable
expressions such as `foo ! bar`.

https://github.com/hashicorp/hcl/blob/main/hclsyntax/parser.go#L1070 seems to suggest this is only
a unary operator.


alisdair (approved):

alisdair:
Thanks!

PR #555, opened 2022-09-22
ext/typeexpr: Convert defaults for optional attrs

alisdair:
When parsing optional object attribute defaults, we previously verified that the default value was convertible to the attribute type. However, we did not keep this converted value.

This commit uses the converted default value, rather than delaying conversion until later. In turn this prevents crashes when transforming collections which contain objects with optional attributes, caused by incompatible object types at the time of defaults application.

Fixes downstream issue hashicorp/terraform#31844


apparentlymart (approved):
This makes sense to me!

PR #563, opened 2022-10-30
Object elements can be separated by comma or newline

bduggan:
Hi!

According to

  https://developer.hashicorp.com/terraform/language/expressions/types#maps-objects

"Key/value pairs can be separated by either a comma or a line break."

But, the documentation here indicates that they may only be separated by a comma.

This is a little change to the docs to include the possibility of separating object elements by a line break.

thanks
Brian

### DIFF
--- a/hclsyntax/spec.md
+++ b/hclsyntax/spec.md
@@ -268,7 +268,7 @@ collection value.
 ```ebnf
 CollectionValue = tuple | object;
 tuple = "[" (
-    (Expression ("," Expression)* ","?)?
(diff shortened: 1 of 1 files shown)


hclsyntax/spec.md
```diff
@@ -271,7 +271,7 @@ tuple = "[" (
     (Expression ("," Expression)* ","?)?
```
kmoe, line 271:
```suggestion
    (Expression (("," | Newline) Expression)* ","?)?
```

kmoe (approved):
I think if we're updating this we'd better do the same for tuples.

Issue #486, opened 2021-08-15
Version v1.0.1-vault-3 and branch 1.0.x-vault have divergent history


maxb:
This release:

https://github.com/hashicorp/hcl/commits/v1.0.1-vault-3

contains two commits which are not included in the branch for this fork:

https://github.com/hashicorp/hcl/commits/1.0.x-vault

and so there is a risk of them being lost in a future release.

(Additionally this branch: https://github.com/hashicorp/hcl/commits/v1.0.1-vault looks like a stale duplicate of the above-mentioned branch.)

maxb:
The situation is now even more confused - a new release `v1.0.1-vault-4` has been made on one side of the divergence, and `v1.0.1-vault-5` has been made on the other!

PR #578, opened 2023-01-02
Reconcile v1.0.1-vault divergent history

maxb:
Relates to #486 

The history of the `v1.0.1-vault-*` releases is somewhat muddled, with divergent history.

This PR makes no tree changes to the `1.0.x-vault` branch, but attaches the history for the releases which are currently not part of the branch, and documents the state of affairs.

The current tagged releases are:

* `v1.0.1-vault` (not actually a tag; instead this is a branch which is used in a tag-like manner, even being present in older Vault `go.mod` files): A backport of #275 for Vault.
* `v1.0.1-vault-2`: A backport of unused key tracking for Vault, but it reverts part of the #275 backport in a commit just titled "wip"
* `v1.0.1-vault-3`: Includes all changes in `v1.0.1-vault-2`, but is not a descendant of the tagged `v1.0.1-vault-2` commit, nor on the current `1.0.x-vault` branch
* `v1.0.1-vault-4`:
  * Squash-merges the changes from `v1.0.1-vault-3` into the `1.0.x-vault` branch
  * Re-introduces the part of #275 lost in `v1.0.1-vault-2` (e80118accb521e47bc5b93104bf46c67d89d2242)
  * "VAULT-8519 - Don't erroneously report nested JSON as unused keys (#559)"
* `v1.0.1-vault-5`: By being tagged on the *source* branch of #559, this tag effectively re-reverts the re-introduction of the part of #275 that was present in `v1.0.1-vault`, absent in `-vault-2` and `-vault-3`, then back again in `vault-4`.

After this PR is merged, there is a needed follow-up action: decide in which direction the reverting and unreverting of part of #275 should actually stay in.

If it is supposed to be gone, actually revert e80118accb521e47bc5b93104bf46c67d89d2242 on branch `1.0.x-vault` so that it doesn't get re-introduced if another version is made.

If it is supposed to be kept, then make a new release, `v1.0.1-vault-6` that includes it again.

----

Finally, for clarity, it would be nice to delete the branch used in a tag-like manner, `v1.0.1-vault`, and replace it with real tag of the same name, pointing at the same commit (cf4c4bf2466d28b8fd0e9953ee5eb7caa3af5418)


maxb:
@VioletHynes As you were recently working in this area to produce the `v1.0.1-vault-4` and `v1.0.1-vault-5` releases, would you be willing to take a look at this?

VioletHynes:
I'll take a look at this, yeah! Thanks for doing this. I noticed during my work that the history was a little confusing and muddled, so thanks for looking at it!

VioletHynes:
One note is that the PR I made doesn't prevent unusedKeys from working, it just prevents erroneous reports for some JSON structures. With regards to the decision for the future, I feel fairly confident that the current behaviour is the correct behaviour, though admittedly I made this change quite a while ago at this point.

VioletHynes (approved):

maxb:
Thanks for approving. Note that as a community member, I don't have permission to merge my own PRs even once approved. If you'd like to merge this, issue #486 will still remain open, and I'll copy over the follow-up actions to be tracked there.

VioletHynes:
Apologies! Merging this slipped my mind. Thanks again for this PR! Merging :)

maxb:
Ah... whoops, sorry, I should have been more explicit, but this PR really needed to be merged in "create a merge commit" mode, not squash-merge, to achieve the desired effect of bringing all of the tags into the history of the branch.

Should I create another PR?

Issue #582, opened 2023-01-24
Incorrect range for incomplete `ObjectConsExpr` in `FunctionCallExpr`


radeksimko:
As can be demonstrated by the following snippet, there seems to be disparity in reporting of the range for an incomplete bare incomplete object notation (`ObjectConsExpr`) and one inside a function (`FunctionCallExpr`).

```go
package main

import (
	"fmt"

	"github.com/hashicorp/hcl/v2"
	"github.com/hashicorp/hcl/v2/hclsyntax"
)

func main() {
	// incomplete object inside function
	src := []byte(`foo({test = })`)
	expr, _ := hclsyntax.ParseExpression(src, "test.hcl", hcl.InitialPos)
	fmt.Printf("inside function: %#v\n", expr.Range())

	// incomplete object standalone
	src = []byte(`{test = }`)
	expr, _ = hclsyntax.ParseExpression(src, "test.hcl", hcl.InitialPos)
	fmt.Printf("standalone: %#v", expr.Range())
}
```
```go
// inside function (End == hcl.InitialPos)
hcl.Range{
	Filename:"test.hcl",
	Start:hcl.Pos{Line:1, Column:1, Byte:0},
	End:hcl.Pos{Line:0, Column:0, Byte:0},
}
// standalone (End correctly reported)
hcl.Range{
	Filename:"test.hcl",
	Start:hcl.Pos{Line:1, Column:1, Byte:0},
	End:hcl.Pos{Line:1, Column:10, Byte:9},
}
```

https://go.dev/play/p/aPjpgC394-W

The reason this is important is because it makes it difficult to enable completion in language server for attribute values (i.e. on the RHS), as it is hard to tell where exactly a given position is.

I am aware that there is probably number of other cases where incomplete configuration is parsed in unexpected ways, but the main reason I'm raising this is because of the disparity between two very similar cases.

PR #588, opened 2023-02-09
hclsyntax: Report correct Range.End for FunctionCall w/ incomplete arg

radeksimko:
Closes https://github.com/hashicorp/hcl/issues/582

Passing tests suggest this isn't breaking anything else, but I'm open to suggestions for any more (careful) changes to avoid any unforeseen consequences.

### DIFF
--- a/hclsyntax/expression_test.go
+++ b/hclsyntax/expression_test.go
@@ -1,8 +1,10 @@
 package hclsyntax
 
 import (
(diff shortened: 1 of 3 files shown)


alisdair (approved):
I don't know how to usefully review the parser_test changes, but the change to parser.go seems logical to me.

Issue #607, opened 2023-04-21
Expose marshaller interfaces for expressions and blocks


ropnop:
Hello - 

I'd like to propose an enhancement that would allow me to use this library and extend with custom functionality in my own projects.

I would like to define 2 marshaler interfaces in the library that I can implement in my own project:

```
// ExpressionMarshaler defines an interface for marshaling a field type as an HCL expression.
type ExpressionMarshaler interface {
	MarshalExpression() (*hclwrite.Expression, error)
}

// BlockMarshaler defines an interface for marshalling a type to an *hclwrite.Block
// It can be used to implement custom hcl block writing for structs and types, including
// when used as fields with the block tag
type BlockMarshaler interface {
	MarshalBlock(blockType string) *hclwrite.Block
}
```

In the [EncodeAsBlock](https://github.com/hashicorp/hcl/blob/main/gohcl/encode.go#L63) and [populateBody](https://github.com/hashicorp/hcl/blob/main/gohcl/encode.go#L51) functions, I'd like to then check if the `interface{}` passed in implements these respective interfaces and if it does, just call the function.

This would allow other consumers of this library to extend/customize the marshalling behavior when using `hclwrite` for custom types and structs.

I have a working fork of this and happy to open a PR, but wanted to start a discussion here first if you are okay with this approach.

PR #608, opened 2023-05-10
Add BlockMarshaler and ExpressionMarshaler interfaces

ropnop:
Hello! This addresses my feature request in #607 . Internally we have a need for this as we want to be able to implement custom types that either emit raw, unescaped HCL, and be able to encode `map[string]string` to HCL easily.

These 2 interfaces let us implement the custom marshalling logic in our code while introducing minimal changes to this library code. 

I added a few tests in `encode_test.go` to demonstrate using these interfaces to overwrite expression and block marshaling behavior on custom types, as well as some tests for handling pointer, nil and zero cases (nil and zero will panic - i think this desired behavior)

Let me know what you think - thanks!

### DIFF
--- a/gohcl/encode.go
+++ b/gohcl/encode.go
@@ -12,6 +12,18 @@ import (
 	"github.com/zclconf/go-cty/cty/gocty"
 )
 
+// ExpressionMarshaler defines an interface for marshaling a field type as an HCL expression.
+type ExpressionMarshaler interface {
+	MarshalExpression() (*hclwrite.Expression, error)
+}
+
+// BlockMarshaler defines an interface for marshalling a type to an *hclwrite.Block
+// It can be used to implement custom hcl block writing for structs and types, including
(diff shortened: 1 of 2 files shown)


yuri-tceretian:
Great feature! That can make encoding to HCL really powerful and let us implement custom encoding the same way we do JSON\YAML marshallers.

It would be great if maintainers could take a look and eventually merge it.

mickeypash:
@ansgarm what are the steps to getting this merged?

crw:
Hi, we discussed this in triage today. This looks like a feature that we would be willing to review. The main barrier is the time it would take to understand why it was not implemented this way originally, to make sure we were not inadvertently introducing new issues by committing this change. Unfortunately, this comes down to maintainer availability. Hopefully the next time someone is doing a deep dive into HCL, we can pick this back up. Thanks!

PR #613, opened 2023-05-28
SetAttribute* methods are not returning new Attribute

jfremy:
The SetAttribute* methods are supposed to return the considered attribute (new or existing) However, because of the variable shadowing in the creation path,  it does not return the new Attribute and instead return nil

Fix the code to return the create Attribute when it is not already present in the body


bbasata:
Done in #806.

PR #620, opened 2023-07-25
Nomad v2.17.0+tweaks

angrycub:
This branch applies all of the Nomad tweaks to the v2.17.0 release. This will enable Nomad and Nomad-pack to leverage to the Extra field on Diagnostics.

### DIFF
--- a/gohcl/decode.go
+++ b/gohcl/decode.go
@@ -14,6 +14,35 @@ import (
 	"github.com/zclconf/go-cty/cty/gocty"
 )
 
+// ExpressionDecoderFunc represents custom expression decoder for a specific type
+type ExpressionDecoderFunc func(expr hcl.Expression, ctx *hcl.EvalContext, val interface{}) hcl.Diagnostics
+
+// BodyDecoderFunc represents custom body decoder for a specific type
+type BodyDecoderFunc func(body hcl.Body, ctx *hcl.EvalContext, val interface{}) hcl.Diagnostics
+
+type Decoder struct {
+	exprConvertors map[reflect.Type]ExpressionDecoderFunc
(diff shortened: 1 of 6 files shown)


gohcl/decode.go
```diff
@@ -31,27 +60,51 @@ import (
...
 	default:
 		panic(fmt.Sprintf("target value must be pointer to struct or map, not %s", et.String()))
 	}
 }
 
-func decodeBodyToStruct(body hcl.Body, ctx *hcl.EvalContext, val reflect.Value) hcl.Diagnostics {
+func (d *Decoder) decodeBodyToStruct(body hcl.Body, ctx *hcl.EvalContext, val reflect.Value) hcl.Diagnostics {
+	if fn, ok := d.bodyConvertors[val.Type()]; ok {
+		return fn(body, ctx, val.Addr().Interface())
+	}
```
angrycub, line 106: Nomad allows for constraint blocks where certain `operators` can be expressed as LHS constants. For example, the canonical `distinct_hosts` constraint is written as follows:

```hcl
constraint {
  operator  = "distinct_hosts"
  value     = "true"
}
```
However, Nomad supports an alternative representation of this same constraint.

```hcl
constraint {
    distinct_hosts = true
}
```
Before re-adding this code, the shortened form would produce the following error.

```text
Unsupported argument; An argument named "distinct_hosts" is not expected here.
```

Fixes https://github.com/hashicorp/nomad/pull/18122#issuecomment-1813537980

tgross:
We're now tagging Nomad-specific releases for HCL1 and HCL2 (see https://github.com/hashicorp/nomad/pull/26089 for more discussion), so this is safe to close.

PR #638, opened 2023-10-24
chore: unnecessary use of fmt.Sprintf

testwill:


crw:
Thanks for this submission! We generally do not accept changes that are typos in comments (or in this case, a no-op use of `fmt.Sprintf`) to reduce `git blame` confusion and merge conflicts in the future. Thanks very much for your submission!

PR #656, opened 2024-02-14
prefer iterator error over for_each

DanielMSchmidt:
If the iterator is misconfigured (e.g. a string instead of a reference) it leads to
follow up issues in the validation of for_each which in turn leads to misleading
error messages. See hashicorp/terraform#34132 for an example.

### DIFF
--- a/ext/dynblock/expand_body_test.go
+++ b/ext/dynblock/expand_body_test.go
@@ -709,3 +709,81 @@ func TestExpandUnknownBodies(t *testing.T) {
 	})
 
 }
+
+func TestExpandInvalidIteratorError(t *testing.T) {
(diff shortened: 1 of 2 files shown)


liamcervante (commented):
Where are the error strings in your description coming from? Is that Terraform doing some additional post processing over the diagnostics returned by HCL?

DanielMSchmidt:
> Is that Terraform doing some additional post processing over the diagnostics returned by HCL?

No, they are added here: https://github.com/hashicorp/hcl/blob/1e1a6b858a91ce49670836e6ba57674d9d183c73/ext/dynblock/expand_spec.go#L44

liamcervante (approved):

PR #659, opened 2024-02-16
Fix license headers

radeksimko:
This aims to fix a number of problems:

 - one static file had a missing header, as surfaced in https://github.com/hashicorp/hcl/pull/637
 - some generated files had license headers added by the `copywrite` tool without reflecting the fact they are generated (and that they should get generated with the headers already)
 - `go generate ./...` implies a dependency on stringer, which is a Go library that can be tracked as such to provide reproducible results.
   - There are actually more dependencies which we cannot track like this but I am just addressing one problem at a time here.
     - Installation of Ragel on Linux/Windows/macOS would be an achievement of its own.
     - That one Ruby script could probably be rewritten into Go rather than us having to depend on Ruby

--- 

Closes https://github.com/hashicorp/hcl/pull/637

--- 

Next step, which I'm expecting to be a separate PR, would be to add some ways of checking for license headers both in the CI and locally, perhaps similar to what we do in `hashicorp/terraform`.


hclsyntax/scan_tokens.go
```diff
@@ -263,7 +262,7 @@ var _hcltok_trans_keys []byte = []byte{
 	233, 234, 237, 239, 240, 243, 48, 57,
 	65, 90, 97, 122, 196, 218, 229, 236,
 	10, 170, 181, 183, 186, 128, 150, 152,
-	182, 184, 255, 192, 255, 128, 255, 173,
+	182, 184, 255, 192, 255, 0, 127, 173,
```
radeksimko, line 265: I'm not entirely sure about the meaning of this byte slice but my understanding is that it will change even with a single character change in the corresponding Ragel file it is generated from.

Since two lines of license headers are being added there, that would explain this.

json/tokentype_string.go
```diff
@@ -4,6 +4,24 @@ package json
 
 import "strconv"
 
+func _() {
```
radeksimko, line 7: This was added back in March 2019 to stringer in https://go-review.googlesource.com/c/tools/+/163637 which implies we have just not run stringer ever since it released the first stable version (`v0.1.0` was released in January 2021).

I avoided pinning to some unstable old version just for that purpose and picked the latest compatible stable version.

radditude (approved):

PR #661, opened 2024-02-16
Add make targets for common checks + fixes

radeksimko:
This is to make it easier for existing and new maintainers and contributors to run checks on the codebase.

These targets can also run from CI which can be implemented as part of a separate PR.

## Examples

```sh
make check
```
```
go run github.com/hashicorp/copywrite@latest headers --plan
Executing in dry-run mode. Rerun without the `--plan` flag to apply changes.

Using license identifier: MPL-2.0
Using copyright holder: HashiCorp, Inc.

Exempting the following search patterns:
hclsyntax/fuzz/testdata/**
hclwrite/fuzz/testdata/**
json/fuzz/testdata/**
specsuite/tests/**

The following files are missing headers:
go vet ./...
"/Users/radeksimko/gopath/src/github.com/hashicorp/hcl/scripts/gofmtcheck.sh"
```

```sh
make fix
```
```
go run github.com/hashicorp/copywrite@latest headers
Using license identifier: MPL-2.0
Using copyright holder: HashiCorp, Inc.

Exempting the following search patterns:
hclsyntax/fuzz/testdata/**
hclwrite/fuzz/testdata/**
json/fuzz/testdata/**
specsuite/tests/**

The following files are missing headers:
2024-02-16T15:25:27.173Z [INFO]  cli: ext/dynblock/options.go modified
gofmt -w ./
```

## Notes

Please note that the license checks specifically are currently failing. That is being fixed in https://github.com/hashicorp/hcl/pull/659

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+fmtcheck:
+	"$(CURDIR)/scripts/gofmtcheck.sh"
+
+fmtfix:
+	gofmt -w ./
+
+vetcheck:
+	go vet ./...
+
+copyrightcheck:
+	go run github.com/hashicorp/copywrite@latest headers --plan
+
+copyrightfix:
+	go run github.com/hashicorp/copywrite@latest headers
+
+check: copyrightcheck vetcheck fmtcheck
+
+fix: copyrightfix fmtfix
--- a/scripts/gofmtcheck.sh
+++ b/scripts/gofmtcheck.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+if [[ -n $(gofmt -l ./) ]]; then echo "Please run gofmt -w ./ to format code"; exit 1; fi;


Makefile
```diff
@@ -0,0 +1,18 @@
+fmtcheck:
```
radditude, line 1: I can't help but notice that this file is itself missing a copyright header 😆

radditude (approved):

Makefile
radeksimko: Good catch! I don't think it matters _at this point_.

The reason is that the tool basically does not know how to add the header to makefiles, it's not that it would be intentionally _ignored_.

See https://github.com/hashicorp/copywrite/blob/0f077594eee1ea09d721a7e6fc9f07302aaff7fe/addlicense/main.go#L360-L392

Also the same issue applies to the upstream project our `copywrite` has forked, although it does have some pending PRs:
 - https://github.com/google/addlicense/pull/147
 - https://github.com/google/addlicense/pull/104
 - https://github.com/google/addlicense/pull/54

To be fair there is support for `*.cmake` but I don't have enough context/knowledge to tell what that is and what's the relationship with `Makefile` or `GNUMakefile` and I am not quite convinced there's a value in learning that just to add the license headers. I would perceive the licensing risk with makefiles in general to be relatively low in comparison with the rest of the codebase.

Feel free to correct me if I'm wrong in any of the above though!

radditude: Agreed, it seems pretty harmless to leave the Makefile un-headered!

PR #673, opened 2024-04-08
Add additional function for parsing traversals with [*] keys

liamcervante:
This PR will eventually be used within Terraform to parse the "partial" addresses used by the deferred actions. We need support for the wildcards within traversals in order to reuse the larger logic for parsing traversals in general.

I've created a separate function for this so that this is a non-breaking change. For now, I haven't updated the behaviour of `SplatTraversal` so the automatic traversal functions will panic if called on a Traversal that contains the `[*]` index. We don't need this behaviour in Terraform (yet?) so I didn't want to introduce anything that couldn't be changed later.

### DIFF
--- a/hclsyntax/parse_traversal_test.go
+++ b/hclsyntax/parse_traversal_test.go
@@ -261,7 +261,9 @@ func TestParseTraversalAbs(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.src, func(t *testing.T) {
 			if test.src == "foo[*]" {
-				// Skip the test that introduces splat syntax.
+				// The foo[*] test will fail because the function we test in
+				// this branch does not support the splat syntax. So we will
+				// skip this test case here.
 				t.Skip("skipping test for unsupported splat syntax")
 			}
 
@@ -282,8 +284,11 @@ func TestParseTraversalAbs(t *testing.T) {
 
 		t.Run(fmt.Sprintf("partial_%s", test.src), func(t *testing.T) {
 			if test.src == "foo[*].bar" {
-				// Skip the test that's supposed to fail for splat syntax.
-				t.Skip("skipping test for unsupported splat syntax")
+				// The foo[*].bar test will fail because the function we test in
+				// this branch does support the splat syntax and this test is
+				// designed to make sure that the other branch still fails with
+				// the splat syntax. So we will skip this test case here.
+				t.Skip("skipping test that fails for splat syntax")
 			}
 
 			got, diags := ParseTraversalPartial([]byte(test.src), "", hcl.Pos{Line: 1, Column: 1})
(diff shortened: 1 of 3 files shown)


hclsyntax/parse_traversal_test.go
```diff
@@ -208,10 +210,61 @@ func TestParseTraversalAbs(t *testing.T) {
...
+			},
+			1,
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.src, func(t *testing.T) {
+			if test.src == "foo[*]" {
+				// Skip the test that introduces splat syntax.
+				t.Skip("skipping test for unsupported splat syntax")
```
DanielMSchmidt, line 265: Do I understand this right that we skip it here since we use `ParseTraversalAbs` in which case this is unsupported, but include it in the partial_ test cases since `ParseTraversalPartial` supports it?

hclsyntax/parse_traversal_test.go
```diff
@@ -226,5 +279,26 @@ func TestParseTraversalAbs(t *testing.T) {
 				}
 			}
 		})
+
+		t.Run(fmt.Sprintf("partial_%s", test.src), func(t *testing.T) {
+			if test.src == "foo[*].bar" {
+				// Skip the test that's supposed to fail for splat syntax.
+				t.Skip("skipping test for unsupported splat syntax")
```
DanielMSchmidt, line 286: Why don't we test for this to error?

DanielMSchmidt (approved):
Looks got to me from my limited understanding

hclsyntax/parse_traversal_test.go
liamcervante: Yeah, exactly. Basically, I wanted to make sure that using the splat was still broken in the old implementation but worked in the new one. I figured the easiest way to do that was to just manually skip the relevant tests since there's only one of each branch that needs to be skipped. 

I could complicate the whole function by adding in some additional metadata to the test cases, but thought it was overkill just for one case.

hclsyntax/parse_traversal_test.go
liamcervante: ah, this is a copy-paste error which makes the message wrong. This test case is expecting an error, which is not what we want here. So I skip this test case in the test branch that does support splats, otherwise the test will fail expecting an error when there isn't one.

liamcervante: I'll fix this shortly!

jbardin (approved):
This LGTM. The behavior is entirely opt-in, and only used from within Terraform, so we should be able to control its use internally.

nfagerlund (approved):
This makes sense to me, I think! 

The "compare test.src and maybe skip" pattern feels like a minor comprehension hazard for our future selves, but I'm probably over-optimizing; if it's only ever gonna be a couple of tests involved, I won't sweat it.

PR #686, opened 2024-07-09
docs(ext/dynblock): recursive function call typo in detecting variables

calmonr:
I'm studying HCL internals and when reading the docs for detecting variables in the `dynblock` package I got confused by the recursive function call but I noticed that for the depth-first, pre-order traversal the function name was wrong.

It's a small adjustment and I hope I can continue contributing more.

Thank you.


kalvarahul (approved):

liamcervante (approved):
Good find! Thanks for raising.

PR #692, opened 2024-08-21
feat: return an `ExprSyntaxError` for invalid references that end in a dot

ansgarm:
Return an `ExprSyntaxError` for invalid references that end in a dot (commonly occurs in editors for completions).

Detect value expressions of the `ExprSyntaxError` type when parsing object constructor expressions and use them to add an item to the result even though we skip parsing the object due to recovery after the invalid expression.

This allows the Terraform language server to support completions for object attributes after a dot was typed.

### DIFF
--- a/hclsyntax/parser_test.go
+++ b/hclsyntax/parser_test.go
@@ -2672,6 +2672,87 @@ block "valid" {}
 				},
 			},
 		},
+		{
+			"a = { b = c. }",
+			1,
+			&Body{
+				Attributes: Attributes{
+					"a": {
+						Name: "a",
+						Expr: &ObjectConsExpr{
+							Items: []ObjectConsItem{
+								{
+									KeyExpr: &ObjectConsKeyExpr{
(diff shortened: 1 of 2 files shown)


hclsyntax/parser_test.go
```diff
@@ -2672,6 +2672,87 @@ block "valid" {}
...
+									},
+									ValueExpr: &ExprSyntaxError{
+										Placeholder: cty.DynamicVal,
+										SrcRange: hcl.Range{
+											Start: hcl.Pos{Line: 1, Column: 11, Byte: 10},
+											End:   hcl.Pos{Line: 1, Column: 13, Byte: 12},
+										},
+										ParseDiags: hcl.Diagnostics{
+											{
+												Severity: 1,
```
cam72cam, line 2710:
```suggestion
												Severity: hcl.DiagError,
```

radeksimko (approved):
This looks reasonable to me! 👍🏻

liamcervante (approved):

PR #699, opened 2024-09-24
Preserve marks when traversing unknown values

jbardin:
When traversing an unknown value or a `DynamicVal`, the marks from that initial value must be preserved for HCL Index and GetAttr operations. This mirrors the behavior of `GetAttr` and `Index` when used directly the underlying cty values.


radeksimko (approved):
LGTM!

PR #701, opened 2024-10-01
docs: fix typo under "Dynamic Attributes Processing"

bschaatsbergen:
Fixes a minor typo to rephrase the beginnign of the pragraph to:

"The _schema-driven_ processing model is useful when the expected structure of a body is known by the calling application."

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -96,7 +96,7 @@ of the implementation language.
 ### _Dynamic Attributes_ Processing
 
 The _schema-driven_ processing model is useful when the expected structure
-of a body is known prior to the calling application. Some blocks are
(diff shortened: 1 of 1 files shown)


crw:
https://www.google.com/search?q=a+priori

> relating to or denoting reasoning or knowledge which proceeds from theoretical deduction rather than from observation or experience.
> "a priori assumptions about human nature"

Probably that phrase should be in italics to denote the latin legalistic meaning. That said, I think if the sentence works as well without it, we should remove it.

crw:
The more I think about it, the more I think _a priori_ this should be removed.

PR #704, opened 2024-10-28
Update V1 Parser to Check for Duplicates 

sarahalsmiller:
There was a consul security finding where when parsing in an ACL object using the V1 parser that looked like the following

```
acl = "read"
acl= "write"
```

The decoder would only only return the last value in the list, causing a potential for accidental privilege escalation. This is a known issue with V1 and has been resolved in V2, however, because the APIs are incompatible and consul has many structs already using the V1 syntax, it didn't seem practical to do a hard shift to V2 HCL at this time. This PR should fix the issue in V1 and add additional test cases.


GreenLantern16:
Can you unsubscribe ***@***.*** from this list? He is no
longer at HashiCorp.

NiniOak (approved):
Tested locally and works to remove duplicate acl configuration. Thanks @sarahalsmiller

finnigja:
> Can you unsubscribe ***@***.*** from this list? He is no longer at HashiCorp.

We can't control your GitHub email notification settings, @GreenLantern16.  I think if you hit the unsubscribe link at the bottom of the email you received, that should take care of it.

GreenLantern16:
Thanks for the response. @GreenLantern16
<https://github.com/GreenLantern16> left
HashiCorp last spring. I receive all of his emails. I will attempt to
contact him about this.

PR #710, opened 2024-11-14
Better control of marks through conditional and for expressions

jbardin:
The unknown value paths within conditional and for expressions were not consistently retaining marks.

In the case of for expressions, an unknown collection would lose all marks. If we have the any marks from the collection we can always return those. If the expression contains a conditional include, we can also attempt to obtain the marks from that since the conditional impacts the total collection value, the marks must alway be included in the final result. This gives us a more complete unknown value, but it can never contain a comprehensive set of marks for all cases. If the collection is entirely unknown, we have no idea what the element value marks are going to be though, so the final known result could still always include marks we were not anticipating.

The conditional expressions are more straightforward to fix, but also more troublesome from Terraform's perspective because it represents a change in behavior. Previously, the marks for the true and false values were kept distinct and were not included at all if the result was unknown. This poses a problem when behavior is specified by marks on the values however, because the the behavior changes depending on whether the result is known or not. The only way to reconcile this with conditionals is to always combine all marks, which is what I've done here. The union of marks definitely makes logical sense when the condition is unknown, since we could have either set, but could be surprising for configurations which could never produce an unknown condition. 

The change in marks looks like so (using letters for variables and numbers for marks):

|  `x(1) ? y(2) : z(3)` | `x` Unknown | `x` True | `x` False
|----|----|----|----|
| before | `<unknown>()` | `y(2)` | `z(3)` |
| after | `<unknown>(1,2,3)` | `y(1,2,3)` | `z(1,2,3)` |

You can see here that if `1` represented "sensitive" as used in Terraform, the result which is derived from a sensitive value has no indication of that.  A case could be made that the y and z values should be `y(1,2)` and `z(1,3)`, but conditional expressions already need to unify the return type, so it feels natural that it should also unify the _meta_-type information like marks, making the results more consistent and predictable.

### DIFF
--- a/hclsyntax/expression.go
+++ b/hclsyntax/expression.go
@@ -788,21 +788,24 @@ func (e *ConditionalExpr) Value(ctx *hcl.EvalContext) (cty.Value, hcl.Diagnostic
 		})
 		return cty.UnknownVal(resultType), diags
 	}
+
+	// Now that we have all three values, collect all the marks for the result.
+	// Since it's possible that a condition value could be unknown, and the
+	// consumer needs to deal with any marks from either branch anyway, we must
+	// always combine them for consistent results.
+	condResult, condResultMarks := condResult.Unmark()
+	trueResult, trueResultMarks := trueResult.Unmark()
+	falseResult, falseResultMarks := falseResult.Unmark()
+	var resMarks []cty.ValueMarks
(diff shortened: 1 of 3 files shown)


radeksimko (approved):
LGTM

PR #719, opened 2024-12-17
fix boolean short-circuit diagnostic handling

jbardin:
Follow up to #713.

A couple boolean expression tests were passing because diagnostics were accidentally dropped when there was no short-circuit return. Add a positive test for the diagnostics and make sure they are always collected from both sides of the expression.

The logic to drop diagnostics needed to be within the short-circuit implementations, where a good LHS unknown value (one without any errors) prevents further evaluation allowing us to optimistically skip the RHS diagnostics. This can be confusing, because the logic looked correct, but we need to remember that cty/hcl return an unknown value when faced with errors too, and part of the short-circuit logic is to selectively handle errors when possible.


mukeshjc (approved):
LGTM. Thank you.

PR #729, opened 2025-01-30
CVE fix by bumping Deps

mohanmanikanta2299:
Updated the golang.org/x/crypto dependency version to fix the CVEs listed below.

- https://hashicorp.atlassian.net/browse/SECVULN-17063
- https://hashicorp.atlassian.net/browse/SECVULN-13803


.github/workflows/checks.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
         with:
-          go-version-file: go.mod
+          go-version: "1.21"
```
sonamtenzin6, line 51: **Question:** Was there any issue with the previous change ?

sonamtenzin6 (approved):

.github/workflows/checks.yml
mohanmanikanta2299: Copywrite repo was updated and it uses toolchain supported by Go version 1.21 or greater.
But the current go version used in the hcl repo is 1.18 which doesn't support toolchain. Hence the update

PR #734, opened 2025-03-05
go.mod: update github.com/mitchellh/go-wordwrap to v1.0.1

thaJeztah:
Update the dependency to use a tagged release; changes since the version used;

- Add go.mod
- Wrap multibyte characters properly
- Do not break on non-breaking space

full diff: https://github.com/mitchellh/go-wordwrap/compare/ad45545899c7...v1.0.1


mohanmanikanta2299 (approved):
LGTM

radeksimko (approved):
It looks like the library is only used to [wrap diagnostic wording](https://github.com/hashicorp/hcl/blob/b2e5121ba4d22d7785c02f0a2965d117faae67ce/diagnostic_text.go#L203), so this should have minimal risk in spite of the major version update.

Thanks for the PR.

PR #735, opened 2025-03-07
Coverage test

KaushikiAnand:
- I have added unit test coverage in the go-multierror.yml workflow
- The unit test coverage change was done to help understand the defects early in lifecycle and come up with a good solution for it.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -35,12 +35,7 @@ jobs:
         uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5.3.0
         with:
           go-version-file: go.mod
-      - name: Go test (Linux/macOS)
-        if: ${{runner.os != 'Windows'}}
-        run: |
-          go test -v -race -coverprofile=coverage.out ./...
-      - name: Go test (Windows)
-        if: ${{runner.os == 'Windows'}}
+      - name: Go test
         run: |
           go test -v -race -coverprofile="coverage.out" ./...
       - name: Upload coverage report
@@ -48,13 +43,9 @@ jobs:
         with:
           path: coverage.out
           name: coverage-report-${{matrix.target}}
-      - name: Display coverage report (Linux/macOS)
+      - name: Display coverage report                   # displayed only for linux and macOS
         if: ${{runner.os != 'Windows'}}
         run: go tool cover -func=coverage.out
-      - name: Display coverage report (Windows)
-        if: ${{runner.os == 'Windows'}}
-        run: go tool cover -func=coverage.out
-        shell: cmd
 
   copyright:
     name: "copyright headers"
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -35,9 +35,26 @@ jobs:
...
-      - name: Go test
+      - name: Go test (Linux/macOS)
+        if: ${{runner.os != 'Windows'}}
         run: |
-          go test ./... -race
+          go test -v -race -coverprofile=coverage.out ./...
+      - name: Go test (Windows)
+        if: ${{runner.os == 'Windows'}}
+        run: |
+          go test -v -race -coverprofile="coverage.out" ./...
```
sonamtenzin6, line 45: Question: Why did we need to add difference checks for windows and other OS, was this not working fine before ?

.github/workflows/checks.yml
```diff
@@ -35,9 +35,26 @@ jobs:
...
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808
+        with:
+          path: coverage.out
+          name: coverage-report-${{matrix.target}}
+      - name: Display coverage report (Linux/macOS)
+        if: ${{runner.os != 'Windows'}}
+        run: go tool cover -func=coverage.out
+      - name: Display coverage report (Windows)
+        if: ${{runner.os == 'Windows'}}
+        run: go tool cover -func=coverage.out
```
sonamtenzin6, line 56: Question: Why do we need different statements here ? Both the run commands are same

.github/workflows/checks.yml
KaushikiAnand: yes on running the cmd "go test -v -race -coverprofile=coverage.out ./...", the coverage.out was not being generated for windows. I found this error while working on go-multierror repo and this solution solved the error. So that is why i wrote different checks for windows and other OS.

.github/workflows/checks.yml
KaushikiAnand: for windows after run cmd i have added a "shell: cmd" as it runs it in a shell env instead of powershell.
When running the this cmd "go tool cover -func=coverage.out", the display of the coverage report was not occurring, so in order to solve this error, i had to write my ci workflow in this way.

.github/workflows/checks.yml
sonamtenzin6: can you check if `go test -v -race -coverprofile="coverage.out" ./...` can be used for other OS ? because adding if conditions in workflow will decrease readability

KaushikiAnand: okay will look into it.

KaushikiAnand: go test -v -race -coverprofile="coverage.out" ./...  this can bee used for other OS also. so changed changed it accordingly.

.github/workflows/checks.yml
KaushikiAnand: suggested changes made

sonamtenzin6 (approved):

PR #739, opened 2025-03-13
Skip marked values when writing diagnostics

wata727:
Fixes https://github.com/hashicorp/hcl/issues/737
Follow up of https://github.com/hashicorp/hcl/pull/738

When the `DiagnosticTextWriter` writes out a diagnostic, if the diagnostic has an Expression and EvalContext, it will also write out the evaluation context in which the diagnostic occurred.

The `cty.Value` of the variable relevant to the expression will be rendered as a string by `valueStr`, but calls to `AsString` or `LengthInt` here will panic on marked values.

It is not clear how HCL should represent such marks, so we skip the marked values from the additional context as a general-purpose diagnostic writer.


crw:
Thanks! Will make sure the team sees this.

PR #750, opened 2025-05-21
test_suite: refactor schema validation of diagnostic file range, pos

abhijeetviswa:
With this refactor, the variables `testFileRangeSchema` and `testFilePosSchema` are no longer unused.

### DIFF
--- a/cmd/hclspecsuite/test_file.go
+++ b/cmd/hclspecsuite/test_file.go
@@ -304,7 +304,7 @@ func (r *Runner) decodeRangeFromBody(body hcl.Body) (hcl.Range, hcl.Body, hcl.Di
 	rangeBody, remain, moreDiags := body.PartialContent(testFileRangeSchema)
 
 	diags = append(diags, moreDiags...)
-	if moreDiags.HasErrors() {
+	if rangeBody == nil {
 		return hcl.Range{}, nil, diags
 	}
 


liamcervante (commented):
Thanks! I agree with the approach, but just some notes about safety and best practices when parsing HCL.

liamcervante (commented):
I think there's still some capacity for panics and/or redundant diagnostics.

The suggestions make it so that each subsequent step is only applied if the preceding step was successful. This makes it so we won't try and access null attributes and we won't test values we already know are invalid.

cmd/hclspecsuite/test_file.go
```diff
@@ -240,36 +240,95 @@ func (r *Runner) decodeDiagnosticsBlock(block *hcl.Block) ([]*TestFileExpectDiag
...
-		Line   int `hcl:"line"`
-		Column int `hcl:"column"`
-		Byte   int `hcl:"byte"`
+func (r *Runner) decodePosFromBody(body hcl.Body) (hcl.Pos, hcl.Diagnostics) {
+	pos := hcl.Pos{}
+	var diags hcl.Diagnostics
+
+	posBody, moreDiags := body.Content(testFilePosSchema)
+	if moreDiags.HasErrors() {
+		diags = append(diags, moreDiags...)
```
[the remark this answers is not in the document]
liamcervante: `.Value()` might return a struct, but `posBody.Attributes["line"].Expr` will panic because `posBody.Attributes` is a map to pointers (I think). So, if you're not returning early you then need to check to make sure the attribute is included in the map otherwise it might panic.

cmd/hclspecsuite/test_file.go
```diff
@@ -240,36 +240,83 @@ func (r *Runner) decodeDiagnosticsBlock(block *hcl.Block) ([]*TestFileExpectDiag
...
+
+func (r *Runner) decodeRangeFromBody(body hcl.Body) (hcl.Range, hcl.Body, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+
+	rangeBody, remain, moreDiags := body.PartialContent(testFileRangeSchema)
+
+	diags = append(diags, moreDiags...)
+	if moreDiags.HasErrors() {
+		return hcl.Range{}, nil, diags
+	}
```
liamcervante, line 300: You don't need this. `rangeBody.Blocks` returns a slice, so even if it's empty it won't crash later.

liamcervante (commented):
Sorry, I forgot to add the extra details in response to your replices 👍

cmd/hclspecsuite/test_file.go
abhijeetviswa: `rangeBody` is a pointer though (`*hcl.BodyContent`).

cmd/hclspecsuite/test_file.go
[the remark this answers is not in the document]
abhijeetviswa: Indeed, `posBody.Attributes` is a map to pointers. 
I've rewritten this function to check for the existence of the attribute to prevent the panic. Thanks!

cmd/hclspecsuite/test_file.go
liamcervante: That's true 👍 You could potentially do something like:

```
diags = append(diags, moreDiags...)
if rangeBody == nil {
  return hcl.Range{}, nil, diags
}
```

That would allow you to keep checking any parts of the configuration that might have been valid (like if the `from` block was parsed successfully, but the `to` block wasn't). I'll leave this up to you though!

liamcervante (approved):
Just one comment left, optional though! Thanks for your patience!

liamcervante (approved):

PR #754, opened 2025-05-23
suppress and fix lint errors by unused

abhijeetviswa:
The majority of unused declarations were one of 3 types:
  - Global constants or sentinel variables. These were untouched becuse they serve as documentation and may become useful in the future.
  - Variable declarations to assert a struct implements an interface. These are compile time safety nets and were also ountouched.
  - Runtime function
  - Unused fields in structs. These were removed to reduce overall since they serve no purpose and at best increase memory consumption and at worst make such structs more confusing to the reader.

There was also an undocumented private function that was removed.  It was straightforward and can be reimplemented if need be.

### DIFF
--- a/hclsyntax/scan_string_lit.go
+++ b/hclsyntax/scan_string_lit.go
@@ -109,8 +109,11 @@ var _hclstrtok_eof_actions []byte = []byte{
 	3,
 }
 
+//nolint:unused
 const hclstrtok_start int = 4
 const hclstrtok_first_final int = 4
+
+//nolint:unused
 const hclstrtok_error int = 0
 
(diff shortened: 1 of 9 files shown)


mukeshjc (dismissed):

hclsyntax/scan_string_lit.go
abhijeetviswa, line 1: This file was regenerated by running `go generate ./hclsyntax/generate.go` and undoes changes made in 314d2366eadcbd243f2988a535587d7fea94442e.

mukeshjc (approved):

liamcervante (approved):

PR #763, opened 2025-07-03
ops: handle unknown objects correctly when looking up by index

liamcervante:
## Description

The linked bug relates to an issue in Terraform where the static checking of objects isn't working as expected. The reason is that objects are converted to unknown values, and then references are walked using the Index function in this repo.

The index function was previously just not validating the requested index at all and returning the object unchanged. For objects we do actually know the attributes even if it is unknown, so we can still validate the index and return an unknown value of the correct type if the index does actually exist.

This fixes the underlying bug in Terraform, since the invalid references will now be properly type checked.

## Related Issue

https://github.com/hashicorp/terraform/issues/37271

## How Has This Been Tested?

Unit tested


mildwonkey (approved):
LGTM!

jbardin (approved):
LGTM!

There seems to be a subtle change in behavior here, where an index could be used on an object to avoid this validation. If for example you have an unknown object with only a foo attribute, 
```
try(object["bar"], true)
```
In Terraform this would change the evaluation from a dynamic `unknown` to `true`, but since resolving an unknown of type `any` to a `bool` is valid, I can't come up with any way this could cause failures in existing config.

Issue #597, opened 2023-03-13
hclsyntax: Recovery of outer `ObjectConsExpr` with incomplete nested `ObjectConsExpr`


radeksimko:
Currently when the configuration contains nested object expressions where the inner one is incomplete, such as

```hcl
attr = {
  foo = {
    bar
  }
}
```

this trips up the parser and results in the `foo` item not being reported at all.

Here's a test to demonstrate the behaviour:

```go
func TestParseConfig_incompleteObject(t *testing.T) {
	tests := []struct {
		input string
		want  *Body
	}{
		{
			`attr = {
  foo = {
    bar
  }
}`,
			&Body{
				Attributes: Attributes{
					"attr": {
						Name: "attr",
						Expr: &ObjectConsExpr{
							Items: []ObjectConsItem{
								{
									KeyExpr: &ObjectConsKeyExpr{
										Wrapped: &ScopeTraversalExpr{
											Traversal: hcl.Traversal{
												hcl.TraverseRoot{
													Name: "foo",
													SrcRange: hcl.Range{
														Start: hcl.Pos{Line: 2, Column: 3, Byte: 11},
														End:   hcl.Pos{Line: 2, Column: 6, Byte: 14},
													},
												},
											},
										},
									},
									ValueExpr: &ObjectConsExpr{
										Items: []ObjectConsItem{},
										SrcRange: hcl.Range{
											Start: hcl.Pos{Line: 2, Column: 9, Byte: 17},
											End:   hcl.Pos{Line: 4, Column: 4, Byte: 30},
										},
										OpenRange: hcl.Range{
											Start: hcl.Pos{Line: 2, Column: 9, Byte: 17},
											End:   hcl.Pos{Line: 2, Column: 9, Byte: 17},
										},
									},
								},
							},
							SrcRange: hcl.Range{
								Start: hcl.Pos{Line: 1, Column: 8, Byte: 7},
								End:   hcl.Pos{Line: 5, Column: 2, Byte: 32},
							},
							OpenRange: hcl.Range{
								Start: hcl.Pos{Line: 1, Column: 8, Byte: 7},
								End:   hcl.Pos{Line: 1, Column: 9, Byte: 8},
							},
						},
						SrcRange: hcl.Range{
							Start: hcl.Pos{Line: 1, Column: 1, Byte: 0},
							End:   hcl.Pos{Line: 5, Column: 2, Byte: 32},
						},
						EqualsRange: hcl.Range{
[...]

PR #785, opened 2026-02-10
add partially complete items in nested object cons structures on error

gotwarlost:
## Description

As issue #597 notes, the nested structure of ObjectCons is not preserved when there is an error in processing the inner object.

This fixes the issue by unconditionally adding the inner item as opposed to only adding it for certain types of errors. This is the exact same processing as is done for tuple cons parsing.

Some unit tests for failure cases  broke as a result and have been updated to the new structures that will be written.

## Context

I've written an [HCL based DSL for processing resources ](https://github.com/crossplane-contrib/function-hcl/blob/main/spec.md) for [Crossplane](https://www.crossplane.io/) use. I'm writing a language server for this.

In this DSL, the resource definition is captured as an object value of a `body` attribute, which means that the entire resource is one object as opposed to Terraform where this is all nested blocks and attributes.

So while object cons recovery is relatively a nice-to-have feature for the Terraform language server given that such structures are few and far-between, my language server is dead in the water without this fix.

Please let me know how I can move this forward.

## Related Issue

Issue #597 

## How Has This Been Tested?

Extensively tested manually for the language server I'm writing. This fix does not change behavior when the HCL is well-formed.

### DIFF
--- a/hclsyntax/parser.go
+++ b/hclsyntax/parser.go
@@ -1441,7 +1441,7 @@ func (p *parser) parseObjectCons() (Expression, hcl.Diagnostics) {
 		}
 
 		// Wrapping parens are not explicitly represented in the AST, but
-		// we want to use them here to disambiguate intepreting a mapping
+		// we want to use them here to disambiguate interpreting a mapping
 		// key as a full expression rather than just a name, and so
 		// we'll remember this was present and use it to force the
 		// behavior of our final ObjectConsKeyExpr.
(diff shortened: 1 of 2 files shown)


zczcmlml:
@jbardin This PR resolves the language server issue in my use case. It would be greatly appreciated if this PR could be merged.

jbardin (approved):
Thanks @gotwarlost!

Issue #365, opened 2020-04-21
Make *AtPos functions work for incomplete config


radeksimko:
### HCL Template
```hcl
provider ""
```

### Parsing logic

```go
package main

import (
	"log"
	
	"github.com/kr/pretty"
	"github.com/hashicorp/hcl/v2"
	"github.com/hashicorp/hcl/v2/hclsyntax"
)

func main() {
	c, diags := hclsyntax.ParseConfig([]byte(`
provider ""
`), "test.tf", hcl.InitialPos)
	if diags.HasErrors() {
		log.Printf("diags: %s", diags.Error())
	}

	log.Printf("config: %s", pretty.Sprint(c))

	b := c.OutermostBlockAtPos(hcl.Pos{
		Line:   2,
		Column: 11,
		Byte:   13,
	})

	log.Printf("block: %s", pretty.Sprint(b))
}
```

### Expected behavior

`OutermostBlockAtPos` in the above example returns the incomplete block.

### Actual behavior

`OutermostBlockAtPos` in the above example returns `nil`.

### Steps to reproduce

`go run` the above code snippet

### Use Cases

When users are editing config and ask the language server via the IDE for completion, it's expectable they will want to complete an incomplete block, such as the above and the language server will therefore need to know where exactly in the incomplete block they're requesting completion.

PR #786, opened 2026-02-16
Set CloseBraceRange to last label in error-recovered blocks

Yantrio:
## Description

When the HCL parser error-recovers from an incomplete block (no opening brace), `Block.Range()` only covered the type keyword, not the labels. This is because `CloseBraceRange` was set to `ident.Range` as a placeholder, so `RangeBetween(TypeRange, CloseBraceRange)` produced a range that ended at the type keyword.

This caused position-based lookups like `OutermostBlockAtPos` to miss positions within labels, which downstream affected things like `completionAtPos` in hcl-lang skipping incomplete blocks entirely and returning snippets instead of real label candidates.

This attempts to fix this by editing the `finishParsingBodyBlock`'s error-recovery path, setting `CloseBraceRange` to the last label range (falling back to `ident.Range` if no labels exist). This ensures that the range includes more information than before.

## Related Issue
- hashicorp/hcl#365
- hashicorp/hcl#708 (Partially)

## How Has This Been Tested?

I've run the entire test suite and included new tests to cover the new functionality.

### DIFF
--- a/hclsyntax/parser.go
+++ b/hclsyntax/parser.go
@@ -234,7 +234,6 @@ func (p *parser) parseSingleAttrBody(end TokenType) (*Body, hcl.Diagnostics) {
 			End:      attr.SrcRange.End,
 		},
 	}, diags
-
 }
 
 func (p *parser) finishParsingBodyAttribute(ident Token, singleLine bool) (Node, hcl.Diagnostics) {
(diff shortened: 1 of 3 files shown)


jbardin (approved):
Thanks @Yantrio!

PR #792, opened 2026-04-15
fix(json): correct 'occured' -> 'occurred' in ParseFile diagnostics

SAY-5:
`json/public.go` raised two diagnostics with the wrong spelling:

```go
Detail: fmt.Sprintf("The file %q was opened, but an error occured while closing it.", filename)  // line 111
Detail: fmt.Sprintf("The file %q was opened, but an error occured while reading it.", filename)  // line 122
```

Both messages are surfaced in HCL diagnostic output to end users when `ParseFile` hits a filesystem error. Replace `occured` with `occurred`. `go build ./...` and `go test ./json/` both pass against current `main`.

### DIFF
--- a/json/public.go
+++ b/json/public.go
@@ -108,7 +108,7 @@ func ParseFile(filename string) (rf *hcl.File, diags hcl.Diagnostics) {
 			diags = append(diags, &hcl.Diagnostic{
 				Severity: hcl.DiagWarning,
 				Summary:  "Failed to close file",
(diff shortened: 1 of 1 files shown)


crw:
Thanks for this submission. If you are interested in having this reviewed, please sign the CLA per https://github.com/hashicorp/hcl/pull/792#issuecomment-4253631198. Note that we do not normally accept PRs that make small changes to text output or comments. However, I will leave this open as occasionally these do get picked up and accepted. Thanks again!

PR #797, opened 2026-05-12
hclsyntax: fix body SrcRange start when file begins with a comment

### LINKED ISSUE #551: Incorrect `srcRange` for files starting with a multiline comment
When a HCL file starts with a multiline comment, the `SrcRange` of the `hclsyntax.Body` starts after the multiline comment.

## Test data

Example file starting with a multiline comment:
```hcl
/********************************
 Some comment
********************************/

data "google_project" "project" {
  project_id = var.project_id
}

```

Example code for parsing the file:
```go
f, _ := hclsyntax.ParseConfig(src, filename, pos)
body, _ := f.Body.(*hclsyntax.Body)
fmt.Println(body.Range())
```

Full example here: https://gist.github.com/dbanck/83e0e7936556eab2184d5d89e9d1feac

## Expected behavior
`SrcRange` for the example file should be `main.tf:1,1-8,1`

## Actual behavior
`SrcRange` for the example file is reported as `main.tf:3,34-8,1`

c-tonneslan:
Fixes #551.

When a file starts with a multiline comment, the top-level body's `SrcRange` was starting at the first non-comment token instead of the beginning of the file.

The root cause: `ParseBody` captures its start position by calling `p.PrevRange()` before reading any tokens. With `NextIndex == 0`, `PrevRange()` fell back to `NextRange()`, which calls `Peek()` and skips comment tokens. So the start range pointed to the first non-comment token, not the file start.

Fix is to return `Tokens[0].Range` directly when no token has been consumed yet. That's the actual first token in the stream, comment or not. If the token slice is empty we still fall back to `NextRange()` as before.

Added a regression test in `TestParseConfig` for a file starting with `/* comment */`.

### DIFF
--- a/hclsyntax/parser_test.go
+++ b/hclsyntax/parser_test.go
@@ -2753,6 +2753,49 @@ block "valid" {}
 				},
 			},
 		},
+
+		// Regression: a file starting with a multiline comment had its body
+		// SrcRange start set to the position after the comment rather than
+		// the beginning of the file.
+		{
+			"/* comment */\na = 1\n",
+			0,
+			&Body{
+				Attributes: Attributes{
+					"a": {
+						Name: "a",
+						Expr: &LiteralValueExpr{
+							Val: cty.NumberIntVal(1),
+							SrcRange: hcl.Range{
+								Start: hcl.Pos{Line: 2, Column: 5, Byte: 18},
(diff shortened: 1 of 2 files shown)


crw:
This PR seems to be fully automated GenAI agent activity. We require pull requests to have human review and ownership of the final product, in this case that seems to be missing. Thanks!

PR #801, opened 2026-05-18
docs(guide): fix broken Go example and nonexistent `gohcl.DecodeBlock` ref in hcldec guide

kiwigitops:
Fixes #800.

Two small doc bugs in `guide/go_decoding_hcldec.rst`:

1. The Partial Decoding Go example referenced an undefined variable `block` inside a loop whose loop variable is `sc`, so the example wouldn't compile. Changed to `sc.Type`.
2. The prose referenced `gohcl.DecodeBlock`, which does not exist in the `gohcl` package (only `DecodeBody` and `DecodeExpression` are public). Changed to `gohcl.DecodeBody`.

```diff
 for _, sc := range c.Services {
-    pluginName := block.Type
+    pluginName := sc.Type
```

```diff
-just as with ``gohcl.DecodeBlock``.
+just as with ``gohcl.DecodeBody``.
```

### DIFF
--- a/guide/go_decoding_hcldec.rst
+++ b/guide/go_decoding_hcldec.rst
@@ -191,7 +191,7 @@ to learn which plugins are needed, but process the block bodies dynamically:
    }
 
    for _, sc := range c.Services {
-       pluginName := block.Type
+       pluginName := sc.Type
 
        // Totally-hypothetical plugin manager (not part of HCL)
        plugin, err := pluginMgr.GetPlugin(pluginName)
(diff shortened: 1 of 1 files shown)


kiwigitops:
@crw I completed the CLA with no errors and then I came back to this and clicked the button to refresh it with the CLA bot and it never updated. Signed CLA on multiple OS and on mobile. pressing the link where it says "Have you signed the CLA already but the status is still pending? [Recheck it](https://cla.hashicorp.com/check/hashicorp/hcl?pullRequest=801)." does not work as intended.

PR #804, opened 2026-06-06
docs: fix hcldec guide example and DecodeBody reference

gajjuu:
## Summary

Fixes two documentation issues in `guide/go_decoding_hcldec.rst`:

* Replaces undefined variable `block.Type` with `sc.Type` in the partial decoding example.
* Replaces nonexistent `gohcl.DecodeBlock` reference with `gohcl.DecodeBody`.

Fixes #800


crw:
Thanks, I will raise this in triage.

PR #807, opened 2026-06-15
hclwrite: export `LeadComments()`/`LineComments()`

bbasata:
## Description

This change exports functions from the `hclwrite` package to read comments.

### DIFF
--- a/hclwrite/ast_block_test.go
+++ b/hclwrite/ast_block_test.go
@@ -488,4 +488,71 @@ func TestBlockSetLabels(t *testing.T) {
 	}
 }
 
-func TestBlockLineComments(t *testing.T) {}
+func TestBlockLeadComments(t *testing.T) {
+	tests := []struct {
+		src         string
+		wantComment string
+	}{
+		{
+			`# block comment
+block "test" {}
+`,
+			"# block comment\n",
+		},
+		{
+			`// block comment
+block "test" {}
+`,
+			"// block comment\n",
+		},
+		{
+			`// block comment
+// that goes on a bit
+// for fun
+block "test" {}
+`,
+			"// block comment\n// that goes on a bit\n// for fun\n",
+		},
+		// Terraform accepts multi-line go-style comments, but hclwrite does not consistently support this style comment.
+		{
+			`/* multiline
+comment
+*/
+block "test" {}
+`,
+			"", // unsupported
+		}, {
+			`/* multiline
+comment
+*/
+block "test" {}
+`,
+			"", // unsupported
+		},
+		{
+			`/* multiline comment, single line */
+block "test" {}
+`,
+			"", // unsupported
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.wantComment, func(t *testing.T) {
+			f := mustParseConfig(test.src)
+			b := f.Body().FirstMatchingBlock("block", []string{"test"})
+			got := string(b.LeadComments().Bytes())
+			if got != test.wantComment {
+				t.Errorf("wrong result\ngot:  %s\nwant: %s", got, test.wantComment)
+			}
+		})
+	}
+}
+
+func mustParseConfig(src string) *File {
+	f, diags := ParseConfig([]byte(src), "", hcl.Pos{Line: 1, Column: 1})
+	if len(diags) != 0 {
+		panic(diags.Error)
+	}
+	return f
+}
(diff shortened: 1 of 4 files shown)


mildwonkey:
I got started writing tests for this and ran into a few things - let me know if any of these are just misunderstandings or testing wrong: 

1. [Go formatted multi-line comments](https://developer.hashicorp.com/terraform/language/syntax/configuration#comments) don't appear to be supported in lead comments, though it's working in line comments so it's very possible I messed up the test case. I can put 

2. Comments are returned with final newlines (maybe fine, this is my vaguest concern!)
I'm not sure if the newlines at the end of the comment are appropriate (newlines within multi-line comments are expected but I _assumed_ - not based on knowledge - that we'd trim the final newline, but that might make sense for comments in a way it doesn't for things like attributes 🤷🏻 ). 

3. Inline comments include a space before the comment indicator:
`test_attribute = foo # comment` returns

``` # comment``` instead of  ```# comment```

3. We lose the last line of (admittedly badly formatted) multi-line inline comments:
Something like this:

```
attr = foo # i am a rather 
                 # long comment, and there's nothing else on this line
                 # so this line here will be returned as the lead line on the next attr, probably? or just get lost.
 other_attr = bar
 ```
 
Just returns the first line (` # i am a rather`)

I'm not sure how we should approach "bad" multi-line inline comments (the ones not using actual multi-line syntax). I don't know if it's realistic to expect hclwrite to tell a multi-line inline comment apart from a single line inline comment + the next attribute's lead comment - is the indentation sufficient? does it have to be exactly indented, or just "more than the attribute is"? Do we even have that info? 

I suspect it's sufficient to say that terraform's behavior is undefined if you use single-line comment indicators for multiline comments (as the comment may get lost or end up with the next attr), but we do need to be deliberate about the choice, and document (and test) the limits. 

If it would be easier, I can push the tests to your branch so you can see the test cases; I'll leave that up to you!

apparentlymart:
FWIW `hclwrite` currently tracks these "special" comment tokens in primarily so that it can take them into account when callers make changes to whatever the comments are attached to.

For example, if a caller deletes an attribute with comments attached to it then `hclwrite` wants to delete the comments at the same time, rather than leaving them behind attached to nothing, or attached to the wrong thing.

The existing rules about which comments get attached and which ones don't were therefore aimed at being good heuristics for which comments an author might expect to be removed along with an item that's being deleted. Other uses for these comments might want slightly different heuristics, but changing the existing rules in-place would cause functions like `Body.RemoveAttribute` to behave slightly differently than they did before.

For that "i am a rather" example, `RemoveAttribute("attr")` would already be doing something arguably "wrong" of course:

```hcl
# long comment, and there's nothing else on this line
# so this line here will be returned as the lead line on the next attr, probably? or just get lost.
other_attr = bar
```

...so the current heuristic isn't good for this situation even when focused on the intended use-case. I'm not sure whether authors actually write comments line this in practice.

mildwonkey:
Indeed, I assumed that the invalid multi-line comments would be unhandled, and left as such. I'm certainly not _mad_ about any of these behaviors, but I wasn't sure what would and wouldn't work, and adding tests mainly to reverse-engineer a feature is always a fascinating way of learning the edge cases. 

I'm certainly not out to break any existing callers, so I updated the tests to match reality as documentation. I'm not worried about many of the odd behaviors (granted, I've generally seen all the things you and @jbardin tend to say people "shouldn't do" out in the wilds - sloppy comments are the least of it 😂)   

@bbasata this is a (very minor) limitation to keep in mind, however, for what you're doing with it.

bbasata:
Chiming in as a consumer of these functions-to-be. tl;dr: all the observations are [1] super helpful and [2] super non-blocking for me. I'm  happy to revisit them in The Future.

> 2. Comments are returned with final newlines (maybe fine, this is my vaguest concern!)

Right on. `SpacesBefore` is an understood consideration when I'm working with hclwrite. Depending on the need, I use `strings.TrimSpace` or (most often) `hclwrite.Format` to tidy the bits for display.

✅

hclwrite/ast_block_test.go
```diff
@@ -487,3 +487,5 @@ func TestBlockSetLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestBlockLineComments(t *testing.T) {}
```
jbardin: Is there supposed to be a test here?

mildwonkey: oh, that's embarrassing, thanks and sorry! tests are in place now!

jbardin (approved):

PR #810, opened 2026-06-17
hclwrite: parse object-construct expressions

bbasata:
## Description

Builds an AST for object-construct expressions such as:
```hcl
a = {
    hat = "derby", (cat) = "calico" }
```

Given an `hclwrite.Expression`, an application can read an object value as:
```go
object := expr.AsObjectConsExpr()
value := object.ValueFor("hat")

tokens := value.Expr().BuildTokens(nil)
```

Retrieval is based on matching a Go string to an HCL identifier. As an implementation choice, there is no way to retrieve the value for the variable expression `(cat)` at the time of this pull request.

https://github.com/hashicorp/hcl/pull/811 is a speculative pull request that builds on this one and introduces the capability to rewrite objects.

## Related Issue

https://github.com/hashicorp/hcl/issues/695 and design notes from Martin in https://github.com/hashicorp/hcl/issues/695#issuecomment-2569776621.

## How Has This Been Tested?

`hclwrite/parser_test.go`

### DIFF
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014, 2025
+// Copyright IBM Corp. 2014, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package hclwrite
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -187,6 +187,18 @@ Traversals:
 	}
 }
 
+func (e *Expression) AsObjectConsExpr() *ObjectConsExpr {
+	var found *ObjectConsExpr
+	e.walkChildNodes(func(n *node) {
+		if o, ok := n.content.(*ObjectConsExpr); ok {
+			found = o
+			return
+		}
+	})
+
+	return found
+}
+
 // Traversal represents a sequence of variable, attribute, and/or index
 // operations.
 type Traversal struct {
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -15,6 +15,7 @@ type Expression struct {
 	inTree
 
 	absTraversals nodeSet
+	wrapped       *node
 }
 
 func newExpression() *Expression {
@@ -22,6 +23,15 @@ func newExpression() *Expression {
 		inTree:        newInTree(),
 		absTraversals: newNodeSet(),
 	}
+}
+
+func (e *Expression) wrap(c nodeContent) {
+	if e.wrapped != nil {
+		panic("this is a problem")
+	}
+
+	e.wrapped = newNode(c)
+	e.children.AppendNode(e.wrapped)
 }
 
 // NewExpressionRaw constructs an expression containing the given raw tokens.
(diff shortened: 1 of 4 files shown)


hclwrite/parser.go
```diff
@@ -376,6 +376,70 @@ func parseBlockLabels(nativeBlock *hclsyntax.Block, from inputTokens) (inputToke
...
+// parseObjectConsExpr parses an object-construct expression, defined as:
+//
+//	object = "{" (
+//	    (objectelem (( "," | Newline) objectelem)* ","?)?
+//	) "}";
+//	objectelem = (Identifier | Expression) ("=" | ":") Expression;
+//
+// It leaves any lead comments and line comments as unstructed tokens.
+//
+// It returns *Expression. The returned *Expression has a child *ObjectConsExpr
```
mildwonkey: outdated comment? It looks like the returned `*Expression` is wrapped in a `*node`

mildwonkey (commented):
This looks reasonable to me! Could you please add a test case with a more complex example? An object with a nested object (or a list, or really anything else)? I already tried one locally, so I know it's working, but it'd be good to have some examples as the parser gets more complex.

```
a = {
      hat = "derby", 
     (cat) = "calico", 
     sprat = { what = "even" }, 
    list = [ "please", "be", "correct", "syntax"] // i didn't try this one so don't quote me
}
```

hclwrite/parser.go
```diff
@@ -376,6 +376,70 @@ func parseBlockLabels(nativeBlock *hclsyntax.Block, from inputTokens) (inputToke
...
+}
+
+// parseObjectConsExpr parses an object-construct expression, defined as:
+//
+//	object = "{" (
+//	    (objectelem (( "," | Newline) objectelem)* ","?)?
+//	) "}";
+//	objectelem = (Identifier | Expression) ("=" | ":") Expression;
+//
+// It leaves any lead comments and line comments as unstructed tokens.
```
jbardin: unstructed?

mildwonkey: I think that's accurate; this is the test output for a comment line:

```
{
   Type: "Tokens",
    Val:  " // a fancy hat\n",
},
```

And Martin's comment that referred to them as such: https://github.com/hashicorp/hcl/issues/695#issuecomment-2569776621

Any suggestions for improvements?

bbasata: "Unstructured" has precedent in hclwrite. `s/unstructed/unstructured/`

bbasata:
```suggestion
// It leaves any lead comments and line comments as unstructured tokens.
```

jbardin (dismissed):

mildwonkey (dismissed):

hclwrite/parser.go
```diff
@@ -396,6 +414,78 @@ func parseExpression(nativeExpr hclsyntax.Expression, from inputTokens) *node {
...
+// An object key is an Identifier or an Expression. In this context, a quoted
+// literal is functionally equivalent to an Identifier.
+func parseObjectConsKeyExpr(nativeExpr *hclsyntax.ObjectConsKeyExpr, from inputTokens) *node {
+	expr := parseExpression(nativeExpr.Wrapped, from)
+	wrapExpr := newObjectConsKeyExpr(expr)
+
+	return newNode(wrapExpr)
+}
+
+// parseObjectConsExpr is specifically interested in string literal expressions
```
magodo, line 475: A typo:

```suggestion
// parseTemplateExpr is specifically interested in string literal expressions
```

bbasata: Thank you ✨

hclwrite/ast_expression.go
```diff
@@ -24,6 +25,15 @@ func newExpression() *Expression {
 	}
 }
 
+func (e *Expression) wrap(c nodeContent) {
+	if e.wrapped != nil {
+		panic("this is a problem")
```
mildwonkey:
```suggestion
		panic("wrap called on already-wrapped Expression")
```

Feel free to change that wording, but the panic message needs a bit more info :)

hclwrite/ast_object_cons_expr.go
```diff
@@ -0,0 +1,95 @@
...
+func (k *ObjectConsKeyExpr) String() string {
+	if k.wrapped == nil {
+		return ""
+	}
+
+	if t, ok := k.wrapped.content.(*Traversal); ok && len(t.steps.List()) > 0 {
+		var b strings.Builder
+		tok := t.steps.List()[0].BuildTokens(nil)
+
+		//nolint:errcheck // strings.Builder returns no error
```
mildwonkey, line 73: Is this accurate? WriteTo may return an error. I don't mind if you leave in the nolint (we're doing the same elsewhere) but I do want to check that the comment is accurate

mildwonkey (commented):
only nitpicks left! the nittiest of picks!

hclwrite/ast_object_cons_expr.go
apparentlymart: All of the write methods of `strings.Builder` are documented as being infallible.

Its `Write` method in particular says:

> Write always returns len(p), nil. 

(where that `nil` is the `error` result.)

mildwonkey (approved):

PR #811, opened 2026-06-17
hclwrite: add read and write API for `ObjectConsExpr`

bbasata:
## Description

Rewrite object values.
```go
config := []byte(`a = {
    hat = "derby", (cat) = "calico" }` + "\n")

bowler := hclwrite.TokensForValue(cty.StringVal(`bowler`))

file, _ := ParseConfig(config, "<inline>", hcl.Pos{})
attr := file.Body().GetAttribute("a")

expr := attr.Expr().AsObjectConsExpr()
key, value := expr.SetItemRaw("hat", bowler)

fmt.Printf("key: ")
key.BuildTokens(nil).WriteTo(os.Stdout)
fmt.Println()
fmt.Println("value: ")
value.BuildTokens(nil).WriteTo(os.Stdout)
fmt.Println()
```

## Related Issue

#810


bbasata:
Rebased onto [2994071](https://github.com/hashicorp/hcl/pull/811/commits/299407189acab786290d8a64cda8b791d3501937). The new commits belonging to this PR are [b5b53ef](https://github.com/hashicorp/hcl/pull/811/commits/b5b53efb491edc88e890991e6c39ce4ba4dd7220) and [321afff](https://github.com/hashicorp/hcl/pull/811/commits/321afffd5bde7d0c4d220f03fa87c6dd330cc874).

bbasata:
> Rebased onto [2994071](https://github.com/hashicorp/hcl/pull/811/commits/299407189acab786290d8a64cda8b791d3501937). The new commits belonging to this PR are [b5b53ef](https://github.com/hashicorp/hcl/pull/811/commits/b5b53efb491edc88e890991e6c39ce4ba4dd7220) and [321afff](https://github.com/hashicorp/hcl/pull/811/commits/321afffd5bde7d0c4d220f03fa87c6dd330cc874).

@mildwonkey: I'm marking this one for review, with no expectation that it's quite ready to merge yet. It's simply the next one in the sequence that we're collaborating on.

I've exercised this in a client application to a high degree of confidence. Now, to turn "the change that works" into "the right change."

hclwrite/ast_object_cons_expr.go
```diff
@@ -0,0 +1,162 @@
...
+	if unwrapped == nil {
+		return ""
+	}
+
+	if ident, ok := unwrapped.content.(*identifier); ok {
+		var b strings.Builder
+		tok := ident.BuildTokens(nil)
+		format(tok) // removes any SpacesBefore
+
+		//nolint:errcheck // FIXME: Propagate errors upward.
```
bbasata: Note: this `nolint` is consistent with existing `hclwrite` behavior. So I view it more as _nota bene_, less as "this is a not-good shortcut."

grep `FIXME` for prior art.

apparentlymart: The write operations of a `strings.Builder` cannot fail, and that's a documented guarantee rather than just a current implementation detail. The `error` return values are there only to satisfy the general-purpose `io.Writer` interface.

bbasata:
> @mildwonkey: I'm marking this one for review, with no expectation that it's quite ready to merge yet. It's simply the next one in the sequence that we're collaborating on.

ℹ️ This new test will see some changes in its tree. This is a consequence of parsing the object-construct item key as an expression, so it now has a richer sub-tree. 
```
TestParse/a_=_{___hat_=_{__style_=_"derby",__color_=_(variable)___},___cat_=_"calico"_}
```

hclwrite/parser.go
```diff
@@ -396,6 +414,105 @@ func parseExpression(nativeExpr hclsyntax.Expression, from inputTokens) *node {
...
 }
 
+// parseObjectConsExpr parses an object-construct expression, defined as:
+//
+//	object = "{" (
+//	    (objectelem (( "," | Newline) objectelem)* ","?)?
+//	) "}";
+//	objectelem = (Identifier | Expression) ("=" | ":") Expression;
+//
+// It leaves any lead comments and line comments as unstructed tokens.
```
bbasata: > unstructed

I tried to rebase. Maybe I will try again.

hclwrite/ast_object_cons_expr.go
bbasata: Yes, that's more correct, thank you :)

hclwrite/unwrap.go
```diff
@@ -0,0 +1,25 @@
+// Copyright IBM Corp. 2014, 2026
```
magodo, line 1: This file seems introduced unintentionally.

=
```diff
@@ -0,0 +1,275 @@
...
+			return nil, nil
+		} else {
+			object.items.Add(object.children.Insert(firstItemNode, item))
+		}
+	}
+	return item.kv()
+}
+
+func (object *ObjectConsExpr) firstItemNode() *node {
+	return object.items.List()[0]
```
magodo, line 155: This can cause panic if there is no item in the `items` (which is what is asserted at the callsite).

=
```diff
@@ -0,0 +1,275 @@
...
+	if item != nil {
+		item.ValueObj().expr.Detach()
+		item.ValueObj().expr = item.ValueObj().children.Append(expr)
+	} else {
+		item = newObjectConsItem()
+		item.init(key, expr)
+		if firstItemNode := object.firstItemNode(); firstItemNode == nil {
+			return nil, nil
+		} else {
+			object.items.Add(object.children.Insert(firstItemNode, item))
```
magodo, line 123: I don't quite get the idea here. Per my understanding we shall append the item to the object (similar to how `SetAttributeRaw` does). In order to do that, just do:

```go
object.items.Add(object.children.Insert(object.children.last, item))
```

(Note that the `object.child.last` is `nil` when there is no items, which means to insert at the begining)

hclwrite/unwrap.go
bbasata: I've removed file `=` in [18b2d4b](https://github.com/hashicorp/hcl/pull/811/commits/18b2d4b88c51e0c15363831239b08ff89727c2f8). Thank you ✨.

Function `unwrapUntilType` is used in `ObjectConsKeyExpr.String`, so I will keep it here.

=
bbasata: Thank you ✨ 

I've now addressed this in 9fe1a5df2ac4

=
bbasata: I will take a closer look. 3 months later, I will need to refresh my memory on this :)

bbasata: I've simplified this in [2e8391e](https://github.com/hashicorp/hcl/pull/811/commits/2e8391e1f81ac7f47dffd2d974239649f9432c8e).

magodo: @bbasata The simplified version seems not correct as it will append the `item` beyond the closing `)`. Instead, we shall make sure to insert it before the closing `)` (just as my version does).

You can verify this by adding a new test case in `TestObjectConsExprSetItemRaw`, e.g.:

```
		{
			`a = {
				hat = "derby", (cat) = "calico" }` + "\n",
			"a",
			"dog",
			Tokens{
				{
					Type:         hclsyntax.TokenOQuote,
					Bytes:        []byte(`"`),
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenQuotedLit,
					Bytes: []byte(`bowler`),
				},
				{
					Type:  hclsyntax.TokenCQuote,
					Bytes: []byte(`"`),
				},
			},
			Tokens{
				{
					Type:  hclsyntax.TokenIdent,
					Bytes: []byte{'a'},
				},
				{
					Type:         hclsyntax.TokenEqual,
					Bytes:        []byte{'='},
					SpacesBefore: 1,
				},
				{
					Type:         hclsyntax.TokenOBrace,
					Bytes:        []byte{'{'},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenNewline,
					Bytes: []byte("\n"),
				},
				{
					Type:         hclsyntax.TokenIdent,
					Bytes:        []byte("hat"),
					SpacesBefore: 2,
				},
				{
					Type:         hclsyntax.TokenEqual,
					Bytes:        []byte{'='},
					SpacesBefore: 1,
				},
				{
					Type:         hclsyntax.TokenOQuote,
					Bytes:        []byte{'"'},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenQuotedLit,
					Bytes: []byte(`derby`),
				},
				{
					Type:  hclsyntax.TokenCQuote,
					Bytes: []byte{'"'},
				},
				{
					Type:  hclsyntax.TokenComma,
					Bytes: []byte{','},
				},
				{
					Type:         hclsyntax.TokenOParen,
					Bytes:        []byte{'('},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenIdent,
					Bytes: []byte("cat"),
				},
				{
					Type:  hclsyntax.TokenCParen,
					Bytes: []byte{')'},
				},
				{
					Type:         hclsyntax.TokenEqual,
					Bytes:        []byte{'='},
					SpacesBefore: 1,
				},
				{
					Type:         hclsyntax.TokenOQuote,
					Bytes:        []byte{'"'},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenQuotedLit,
					Bytes: []byte("calico"),
				},
				{
					Type:  hclsyntax.TokenCQuote,
					Bytes: []byte{'"'},
				},
				{
					Type:  hclsyntax.TokenNewline,
					Bytes: []byte{'\n'},
				},
				{
					Type:  hclsyntax.TokenIdent,
					Bytes: []byte("dog"),
				},
				{
					Type:         hclsyntax.TokenEqual,
					Bytes:        []byte("="),
					SpacesBefore: 1,
				},
				{
					Type:         hclsyntax.TokenOQuote,
					Bytes:        []byte{'"'},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenQuotedLit,
					Bytes: []byte("bowler"),
				},
				{
					Type:  hclsyntax.TokenCQuote,
					Bytes: []byte{'"'},
				},
				{
					Type:         hclsyntax.TokenCBrace,
					Bytes:        []byte{'}'},
					SpacesBefore: 1,
				},
				{
					Type:  hclsyntax.TokenNewline,
					Bytes: []byte("\n"),
				},
				{
					Type:         hclsyntax.TokenEOF,
					Bytes:        []byte{},
					SpacesBefore: 0,
				},
			},
		},
```

bbasata: @magodo: you're right. Fixed in [3aa371e](https://github.com/hashicorp/hcl/pull/811/commits/3aa371e2420efd7d35d3fde1e3fc54def5deae23).

bbasata:
@magodo: I've added `ObjectConsExpr.SetItem` in [1fc827c](https://github.com/hashicorp/hcl/pull/811/commits/1fc827c8f8f8d7f037409c589e86c1e2e1b639a8) and have refactored `SetItem*` accordingly.

magodo:
@bbasata Thanks! I think whether we want to go for `SetItem()` or not depends on which solution to choose for #835.

hclwrite/ast_object_cons_expr.go
```diff
@@ -45,11 +144,54 @@ func newObjectConsItem() *ObjectConsItem {
...
+
+	item.children.AppendUnstructuredTokens(Tokens{
+		{
+			Type:  hclsyntax.TokenEqual,
+			Bytes: []byte{'='},
+		},
+	})
+
+	item.value = item.children.Append(newObjectConsValue())
+	item.ValueObj().children.Append(value)
```
magodo:
```suggestion
	item.ValueObj().expr = item.ValueObj().children.Append(value)
```

Otherwise, setting a new item the second time will cause panic as the `expr` is nil.

bbasata:
@mildwonkey: I've rebased this "stacked" PR onto `main`. Ready for review.

hclwrite/ast_object_cons_expr.go
bbasata: Thank you :) Now fixed in [6df1fe0](https://github.com/hashicorp/hcl/pull/811/commits/6df1fe0327517ceb9f335ffa42beba587e7a7bb6).

hclwrite/ast_object_cons_expr.go
```diff
@@ -18,21 +27,111 @@ func newObjectConsExpr() *ObjectConsExpr {
...
-				return item.value.content.(*ObjectConsValue)
+				return item
 			}
 		}
 	}
 
 	return nil
 }
 
+// ItemFor finds an item that matches the given key and returns the value.
```
mildwonkey:
```suggestion
// ValueFor finds an item that matches the given key and returns the value.
```

hclwrite/ast_object_cons_expr.go
```diff
@@ -45,12 +144,56 @@ func newObjectConsItem() *ObjectConsItem {
...
+	item.children.AppendUnstructuredTokens(Tokens{
+		{
+			Type:  hclsyntax.TokenNewline,
+			Bytes: []byte{'\n'},
+		},
+	})
+	identifier := newIdentifier(newIdentToken(key))
+	keyExpr := newObjectConsKeyExpr(newNode(identifier))
+	item.key = item.children.Append(keyExpr)
+	// item.KeyObj().children.Append(newIdentifier(newIdentToken(key)))
```
mildwonkey: Can this comment go?

hclwrite/ast_object_cons_expr.go
```diff
@@ -45,12 +144,56 @@ func newObjectConsItem() *ObjectConsItem {
 	}
 }
 
+func (item *ObjectConsItem) KeyObj() *ObjectConsKeyExpr {
```
mildwonkey, line 145: I have a confusingly strong feeling that these should be ObjKey and ObjVal (and equally strong feelings that this is a weird thing to nitpick and I'm not sure why I feel so strongly). This isn't a blocking request, feel free to ignore this one.

hclwrite/ast_object_cons_expr.go
```diff
@@ -61,14 +204,22 @@ func newObjectConsKeyExpr(wrapped *node) *ObjectConsKeyExpr {
...
 
+// String returns the name of the object item key, if it is specified by
+// an identifier. Otherwise, it returns an empty string.
 func (k *ObjectConsKeyExpr) String() string {
 	if k.wrapped == nil {
 		return ""
 	}
 
-	if t, ok := k.wrapped.content.(*Traversal); ok && len(t.steps.List()) > 0 {
+	unwrapped := unwrapUntilType(k.wrapped, reflect.TypeFor[*identifier]())
```
mildwonkey: I have a knee-jerk objection to using reflect outside of tests and desperation (fully admitting that sometimes it's the best/right/only choice!); wdyt?

mildwonkey (commented):
I left some minor nitpicks - you can definitely ignore or disagree them as you see fit, nothing is blocking.

magodo:
@mildwonkey Please be noted about this, which might impact your review order:

> @bbasata Thanks! I think whether we want to go for `SetItem()` or not depends on which solution to choose for #835.

hclwrite/ast_object_cons_expr.go
```diff
@@ -61,14 +204,22 @@ func newObjectConsKeyExpr(wrapped *node) *ObjectConsKeyExpr {
...
+	}
+
+	if ident, ok := unwrapped.content.(*identifier); ok {
 		var b strings.Builder
-		tok := t.steps.List()[0].BuildTokens(nil)
+		tok := ident.BuildTokens(nil)
+		format(tok) // removes any SpacesBefore
 
 		//nolint:errcheck // strings.Builder returns no error
 		tok.WriteTo(&b)
```
jbardin: The existing code in this section was committed as WIP, was this supposed to be here? Either way the new commit definitely is in question, I'm pretty sure your String() method is modifying the live tokens in place

hclwrite/ast_object_cons_expr.go
```diff
@@ -18,21 +27,111 @@ func newObjectConsExpr() *ObjectConsExpr {
...
+	}
+}
+
+// SetItemRaw either replaces the expression of an existing item of the given
+// name or adds a new item definition to the end of the object, using the given
+// tokens verbatim as the expression.
+//
+// The same caveats apply to this function as for NewExpressionRaw on which it
+// is based. If possible, prefer to use SetItemValue or SetItemTraversal.
+func (object *ObjectConsExpr) SetItem(key string, expr *Expression) (*ObjectConsKeyExpr, *ObjectConsValue) {
```
jbardin, line 87: Now that we can set items from expressions, there's the problem of a user being able to alias nodes set in multiple places at the same time. I'm not sure if we can easily guard against that, or we should just document that the expression must be newly constructed for this method.

hclwrite/ast_object_cons_expr.go
```diff
@@ -18,21 +27,111 @@ func newObjectConsExpr() *ObjectConsExpr {
...
+// however, ItemFor cannot locate the cat.
+func (o *ObjectConsExpr) ValueFor(key string) *ObjectConsValue {
+	if item := o.ItemFor(key); item == nil {
+		return nil
+	} else {
+		return item.ValueObj()
+	}
+}
+
+// SetItemRaw either replaces the expression of an existing item of the given
```
jbardin: wrong doc comment here

hclwrite/ast_object_cons_expr.go
jbardin: Agree, if we need operate on a type, the function should probably use type parameters instead.

hclwrite/ast_object_cons_expr.go
apparentlymart: The `node.assertUnattached` function was intended to help with writing guarding expressions to avoid this sort of aliasing, though I don't know if it's applicable here.

hclwrite/parser.go
```diff
@@ -466,10 +466,37 @@ func parseObjectConsExpr(nativeExpr *hclsyntax.ObjectConsExpr, from inputTokens)
...
-	expr := parseExpression(nativeExpr.Wrapped, from)
-	wrapExpr := newObjectConsKeyExpr(expr)
+	var objectConsKeyExpr *ObjectConsKeyExpr
+
+	switch wrapped := nativeExpr.Wrapped.(type) {
+
+	// a = {
+	//        hat = "derby" }
+	// appends ObjectConsKeyExpr => identifier
+	case *hclsyntax.ScopeTraversalExpr:
```
jbardin, line 476: Can we add some validation and round-trip tests to this? While not technically valid, you are now accepting traversals like `{ a.b.c = 1 }` and silently rewriting as `{ a = 1 }`.  The former will fail to parse in the other direction of course, but we're hiding that error by changing the input into something else.

hclwrite/ast_object_cons_expr.go
bbasata: I will take a fresh look at it.

I'll also note a deliberate reason why this new `hclwrite` unwrap is different from `hcl.UnwrapExpressionUntil`. The latter allows for nice type assertions because `hcl.Expression` is an interface. Example: https://github.com/hashicorp/hcl/blob/4932c1452af6bd71e2241ba244f253643d1828da/expr_call.go#L22-L25

By contrast, `hclwrite.Node` is a concrete type, and each unwrap needs to navigate from node to content (an interface).

bbasata:
> @mildwonkey Please be noted about this, which might impact your review order:
> 
> > @bbasata Thanks! I think whether we want to go for `SetItem()` or not depends on which solution to choose for #835.

If we're happy with `SetItem()` in this PR -- I am happy with it 😄 -- then we do not need to block on review of #835.

Also note to self: need to add changelogs for new exported functions in a follow-up PR.

hclwrite/ast_object_cons_expr.go
bbasata: Fixed in 763db9a54e1202d95c593f9f85c79a27d790f42b.

hclwrite/ast_object_cons_expr.go
bbasata: Fixed in 763db9a54e1202d95c593f9f85c79a27d790f42b.

hclwrite/ast_object_cons_expr.go
bbasata: Oh! I see the ambiguity now.

I read this as "the key object" and "the value object."

They are also "the object item's key" and "the object item's value."

 I forget the exact motivation at the time of writing. I might have been making it explicit that the key is more ... interesting ... than a plain string.

On a fresh reading, I'm happy to simply say `Key` and `Value`. 👍 🫳 ?

hclwrite/ast_object_cons_expr.go
bbasata: Got it: 5f2edf0b802ca4e0ce0535b402c7dc4f5a03bbc2. 👍 🫳 ?

hclwrite/ast_object_cons_expr.go
bbasata: Agreed this will modify the tokens in-place, which is not the intention. Will revise.

bbasata: Revised in f447519566a244902a604d99beda3063b1c451e3.

hclwrite/ast_object_cons_expr.go
bbasata: `node.assertUnattached` does feel right for this.

hclwrite/parser.go
bbasata: Acknowledged. And we can't fail here -- I expect that would change existing behavior of: parse the not-technically-valid expression, opaquely.

Maybe the desired logic is: if it's a scope traversal with more than 1 token, then parse it as an opaque expression (`parseAnyExpression`). 👍 🫳 ?

hclwrite/ast_object_cons_expr.go
bbasata: Gone. [f84367e](https://github.com/hashicorp/hcl/pull/811/commits/f84367e4be9ed1e90e911e1e9a8c070a7b39c1b4).

hclwrite/ast_object_cons_expr.go
mildwonkey: I like those but I'm also still okay with no changes! sometimes the best answer is "read the method comment if you're not sure" esp. when we all have different default assumptions 😂

bbasata: > I like those but I'm also still okay with no changes! sometimes the best answer is "read the method comment if you're not sure" esp. when we all have different default assumptions 😂

All good! With fresh eyes, I've renamed them to `KeyExpr` and `ValueExpr` for consistency:

* These match the respective field names of `hclsyntax.ObjectConsItem`
* This avoids overloading the name `Value`; in `hclsyntax`, `Value` evaluates an expression and yields a `cty.Value`

hclwrite/parser.go
bbasata: Behavior updated to correct this in d35af181c1b9ba80bd8b752d15ec89d55bb4139f.

Also, TIL: hcl parses a two-step traversal as an object-construct key with no complaint. This commit will keep hclwrite consistent with this behavior.

```console
$ hclfmt -check new-blocks.hcl && echo "ok"
cat {
  what = {
    hat.bat = "mat"
  }
}
ok
```

hclwrite/ast_object_cons_expr.go
mildwonkey: @bbasata sorry, it's not entirely clear - were you going to make a change based on this comment?

hclwrite/parser.go
apparentlymart: The handling of what can appear on the left side of `=` in an object constructor expression is quite subtle, because it wants to be more permissive when doing static analysis than when doing evaluation.

For example in Terraform module calls you can write an argument like this:

```hcl
  providers = {
    aws.foo = aws.bar
  }
```

...and that works because Terraform (at least, the last time I saw the relevant code) uses [`hcl.ExprMap`](https://pkg.go.dev/github.com/hashicorp/hcl/v2#ExprMap) and then calls [`hcl.AbsTraversalForExpr`](https://pkg.go.dev/github.com/hashicorp/hcl/v2#AbsTraversalForExpr) on each of the keys, so none of these traversals ever actually get evaluated in the usual sense.

But if you actually tried to evaluate that expression in the normal way using `Expression.Value`, it would _only at that point_ be considered invalid, due to the additional check here:

https://github.com/hashicorp/hcl/blob/4932c1452af6bd71e2241ba244f253643d1828da/hclsyntax/expression.go#L1313-L1337

Handling this weird dissonance between regular evaluation and static analysis is the main reason why `ObjectConsKeyExpr` exists, but it means that someone who is intending to construct an expression for evaluation would need to generate something shaped like the following instead in order to get the appropriate parse tree for that to actually work:

```hcl
  providers = {
    (aws.foo) = aws.bar
  }
```

hclwrite/ast_object_cons_expr.go
bbasata: > @bbasata sorry, it's not entirely clear - were you going to make a change based on this comment?

Thanks for nudging this forward. I've been a bit stuck on this decision. Specifically, I noticed the precedent that `Body.AppendBlock(*Block)` documents "must not be already attached" but does not assert it.

On a fresh look today, I feel that assertion is fine for `ObjectConsExpr.SetItem`, especially as a newly-exported function.

bbasata: > On a fresh look today, I feel that assertion is fine for `ObjectConsExpr.SetItem`, especially as a newly-exported function.

Done in 502d086c3217.

apparentlymart: I can say that my _intention_ had certainly been to prohibit attaching the same node to multiple different parts of the tree -- it's supposed to be a tree, not an acyclic graph -- so if there are codepaths that permit that without panicking then that's probably an oversight rather than intentional, particularly if the docs say that it isn't allowed even though the implementation doesn't check it.

PR #814, opened 2026-07-07
Prepare for Unicode 17 support in Go 1.27

apparentlymart:
## Description

The forthcoming Go 1.27 will update to Unicode 17 tables and algorithms for the standard library. For consistency with that, this PR prepares HCL to use Unicode 17 for its own tables and in the upstream Unicode-related libraries it depends on. In summary, this deals with everything described in [How Terraform uses Unicode](https://github.com/hashicorp/terraform/blob/144d42f23d8b4fedd787724ae7ff50fbfba5ea04/docs/unicode.md), except for finally updating Terraform's own `go.mod`.

To avoid needing to tightly coordinate switching library versions and Go toolchain versions at the same time, here I've matched the new strategy being used by upstream go-cty where it uses conditional compilation to use Unicode 15 text segmentation on Go 1.25 and 1.26, but automatically switch to Unicode 17 when compiling with Go 1.27 or later. This means that Terraform and other HCL dependents can upgrade to a new HCL version containing these changes without immediately adopting Unicode 17 as long as they remain on Go 1.25 or 1.26, with one exception: the identifier rules.

While preparing this I noticed that the data for `ID_Start` and `ID_Continue` had not been updated for a long while and was still using the Unicode 9 range tables. While addressing that I also noticed that the Ruby script for generating the Ragel file from that data was no longer working on recent Ruby versions, and so I applied the same small fix I previously applied to another copy of that same script in my upstream library go-textseg.

Due to how the code generation in this repository is implemented it's not really feasible to support multiple different versions of the Unicode Identifier specification ([UAX#31](https://www.unicode.org/reports/tr31/)) with conditional compilation and so for that one in particular I made the compromise of unconditionally using Unicode 17 regardless of Go version. I expect this is an okay compromise because it just means that a few more characters will be accepted in HCL identifiers, without changing the treatment of any currently-valid HCL configuration.

You'll see in the diff that I also copied the function previously known as `textseg.TokenCount` into the new `unicodeutil` internal package and simplified it to only support grapheme cluster counting since that's all HCL needs. This is because that function no longer exists in upstream textseg as of v17. That function is so trivial that I doubt it is copyrightable at all, but if it _is_ then I'm the copyright holder and I'm contributing it under the MPL for inclusion in HCL.

## How Has This Been Tested?

I ran the full unit test suite both on Go 1.25 (the current minimum supported version) and on Go 1.27rc1 as a placeholder for the forthcoming Go 1.27 release.


internal/unicodeutil/graphemes_v17.go
```diff
@@ -0,0 +1,14 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+//go:build go1.27
```
radeksimko, line 4: Can you just confirm my understanding here? AFAICT this implies later versions as well.

Meaning ...

If no further Unicode updates and breaking changes are required then no further changes are needed here as this file will get picked up automatically by 1.28, 1.29 etc.

If (hypothetically) we need to make similar changes in the future and want to have scanner codepath exclusive to Go 1.28 we would be:

 - adding another file with `//go:build go1.28`
 - appending ` && !go1.28` to this line

radeksimko (approved):
LGTM, just one in-line question but overall I'm happy to merge.

radeksimko (commented):
One more question - Can you clarify which version of Ruby did you run the Ruby script against or help me understand why I'm still seeing a difference after running `go generate`?

My version
```
ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.arm64e-darwin25]
```

Here's the diff: https://gist.github.com/radeksimko/dc7022ceb2230f10b9d791a62c490d83

internal/unicodeutil/graphemes_v17.go
apparentlymart: Yes, a build tag like this means the specified version _or later_, and so `go1.27` means "1.27 or later", and conversely `!go1.27` means "earlier than 1.27".

Whether later Go versions require updates like this depends on whether they adopt new versions of Unicode. If not then this current form of the code will keep using Unicode 17 for later versions of Go.

apparentlymart:
Was there also a change in one of the Ragel input files alongside that change to the generated code?

I ask because that diff suggests that the Ragel-generated state machine ended up a slightly different shape, but I think the identifiers chosen here are arbitrary and we'd need to look at the input file to understand the meaning of this difference. I'm not sure if this suggests that the upstream Unicode character database changed somehow (which is not something I would expect), that the Ruby script has nondeterministic behavior for certain input, or something else. 🤔

I'm not at the computer where I ran this right now so I can't immediately answer the question about Ruby version but I'll look into that a little later. I'm just wondering in the meantime if a diff to the Ragel input file might help explain what's going on.

radeksimko:
> Was there also a change in one of the Ragel input files alongside that change to the generated code?

No changes there from what I can tell.

I ran the `go generate` again (after deleting your branch locally and checking out fresh though I can't tell if that was reason for a change) and the diff in that file was reduced to basically just line numbers in comments.

I'm not quite sure what line numbers is it _supposed to_ refer to or if these comments even have any function at all. I'm guessing not, so it's safe to ignore and "just weird"?

Here's the full diff: https://gist.github.com/radeksimko/3b3848a04cffa3460a7c2054bb4cdd1d I suspect some of it might be potentially a difference in stringer version but it looks close enough that I'm fine with merging it now. Though of course I'd feel even more confident if there was no diff. 😅

radeksimko:
Just to add some more context in hope we can find what's behind the diff:

**`go version`**
```
go version go1.25.13 darwin/arm64
```

**`ragel --version`**
```
Ragel State Machine Compiler version 6.11 July 2026
Copyright (c) 2001-2009 by Adrian Thurston
```

Let me know if yours differ in any way. I don't mind upgrading or downgrading any of these.

Edit: Also `stringer` should be versioned in the project as part of `golang.org/x/tools` which is pinned to `v0.38.0` in `go.mod` so that can't be it.

apparentlymart:
Ragel tries to use [line directives](https://pkg.go.dev/cmd/compile#hdr-Line_Directives) to make the Go compiler report source ranges inside the `.rl` file for the code that's copied literally from the Ragel input, like here:

https://github.com/hashicorp/hcl/blob/2a3b67226841b8c4c2132068c8fb74947fe2e0b3/hclsyntax/scan_tokens.go#L4220-L4232

...and so it also uses `//line` comments referring to the generated `scan_tokens.go` file to _avoid_ reporting the generated code as being in the `.rl` file.

I can't explain why those line numbers changed without any other changes to the file, particularly since the diff suggests that the old line numbers were correct. For example:

```diff
@@ -12,7 +12,7 @@ import (
	 
 // This file is generated from scan_tokens.rl. DO NOT EDIT.
	 
-//line scan_tokens.go:15
+//line scan_tokens.go:17
 var _hcltok_actions []byte = []byte{
 	0, 1, 0, 1, 1, 1, 3, 1, 4,
 	1, 7, 1, 8, 1, 9, 1, 10,
```

The `-12` and `+12` in the hunk header suggest that this hunk begins at line twelve both before and after the diff. If you count the following lines starting at 12, the old `//line scan_tokens.go:15` and new `//line scan_tokens.go:17` lines _are_ on line 15, so I don't understand why Ragel wanted to report that as being line 17 now for you. It's quite hard to follow the git history for Ragel because it's been moved into and out of "Colm Suite" various times by now, so I wasn't able to confirm if this was perhaps a bug fixed in a different Ragel release, or an intentional change, or something else.

This is the Ragel compiler version I had installed when I ran `go generate` for this PR, though:

```
Ragel State Machine Compiler version 6.10 March 2017
Copyright (c) 2001-2009 by Adrian Thurston
```

Line directives are only used if the Go compiler reports an error somewhere in this file though, so indeed it shouldn't have any runtime effect.

---

The first diff you reported where the state machine was slightly different _could_ have a runtime effect, so I'm glad that's gone now although indeed it would've been nice to be able to explain why it happened. :thinking:

For whatever it's still worth, the Ruby on the computer where I ran `go generate` is:

```
ruby 3.2.3 (2024-01-18 revision 52bb2ac0a6) [x86_64-linux-gnu]
```

---

The `hclsyntax/token_type_string.go` diff does indeed look like you're running a different version of `stringer`.

I think the explanation for that one is easier. It looks like dependabot updated the `golang.org/x/tools` module in a30095a319c212ea004672baa1bb92f6a0da6402, but nobody ran `go generate` after that so the `token_type_string.go` was still whatever the old version generated.

I don't recall now because it was nearly two months ago :grinning: but guess I probably ran `go generate` and then only kept the changes to `unicode_derived.rl` and `scan_tokens.go` because they were the only ones relevant to what I was trying to achieve with this PR.

...so I think the `token_type_string.go` you've generated here is what _ought_ to have been done as part of https://github.com/hashicorp/hcl/pull/774, but better late than never!

radeksimko (changes requested):
I have asked another team mate to run `go generate` on their machine and it also resulted in a diff against this PR, more similar to my first diff, i.e. not just line numbers, so I'm afraid we'll need to get to the bottom of this before merging.

I believe a good start would be having an environment in which we get a reproducible result before diving into what is right or wrong. Perhaps it could just be a GitHub Action workflow which we keep for future testing of every PR in this repo. I'll see if I can set something up.

radeksimko:
Having done some testing I now have enough evidence to suggest that Ragel produces non-deterministic output.

The slightly good news is that the output appears to be OS-dependent and it seems **stable enough on linux/amd64**. I didn't check other Linux architectures. It can produce different output half of the time on darwin/arm64 and different output every time on windows/amd64. As far as I can tell the platform dependency is not intentional.

So I am going to propose that we use output from Linux as canonical for the time being.

To make that easier, I'll raise some PRs to introduce CI workflows which along with a failure also produce a `patch` file which can be applied by the PR author, without them having to actually run Ragel locally on Linux. That should allow us to move forward with this PR and feed this back upstream to Ragel as well.

I hope this proposal makes sense?

apparentlymart:
Thanks for testing that!

Non-deterministic output is certainly concerning, and the variation by OS in something I would've expected to be pure computation work is particularly odd.

What you proposed seems pragmatic, though. Perhaps in the long run moving away from Ragel is the best outcome but that would presumably be a large project and so I would be glad for a workaround that avoids blocking on a big rewrite of the scanner.

If it isn't even deterministic on Linux then that would make the check you described annoying for those doing unrelated work, but it seems like you're confident that it will behave consistently enough on Linux in GitHub Actions.

radeksimko:
> Perhaps in the long run moving away from Ragel is the best outcome but that would presumably be a large project and so I would be glad for a workaround that avoids blocking on a big rewrite of the scanner.

I agree this is unlikely to happen in the near future.

Out of my own curiosity (and in the context of https://github.com/hashicorp/hcl/commit/d57901de5f3ea244dcba09683306cac9afb25af2 😄), are you implying hand-written scanner is the alternative here or any particular different solution?

apparentlymart:
I have no specific ideas for what would replace Ragel. At the time of that commit I did some research of what was available and found Ragel the best available compromise.

The situation is probably different now and so if it were up to me I'd start by doing a fresh survey of the available options, including the possibility of a fresh attempt at a hand-written scanner[^1], or at having a language model generate one, along with checking out some of the many new Go-native parser generators that folks have written in the meantime.

[^1]: ...which can now be evaluated against the concrete grammar that exists rather than the the potential grammar I was imagining at the time of original work.

radeksimko:
@apparentlymart Now that https://github.com/hashicorp/hcl/pull/834 was merged, could you please rebase this PR and also bump the Ruby version as part of this PR? Assuming you use Linux locally, there should be hopefully no diff reported by the CI.

apparentlymart:
I have rebased this, which included re-running the generator locally since the generated files had become conflicted by the other recent work.

A Ruby upgrade is not actually necessary for this change: the change I made in `unicode2ragel.rb` is to no longer rely on a "magic" feature that was removed in later Ruby versions, but the non-magic variant works in older versions (including 2.7) too.

Therefore I'm intending to leave it with the currently-selected version for now with the assumption that someone will consider a Ruby upgrade in a separate PR later. I considered removing the `unicode2ragel.rb` change from this PR completely, but I concluded it's better to leave it in just so there's one less maintenance hazard for someone else who might try to upgrade Ruby later, since we've already seen that updated version work on both Ruby 2.7 and Ruby 3.2.

radeksimko (approved):

PR #826, opened 2026-08-20
Add security model documentation

james-warren0:
## Description

This adds an explicit security model that outlines some of the responsibilities for the library, as well as those deferred to integrators of the library.

## Related Issue

N/A

## How Has This Been Tested?

N/A

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,60 @@
+# Security Model
+
+HCL is a toolkit for building structured, expressive configuration languages. The
+evaluator is powerful by design: it resolves variables, calls functions, and processes
+complex expressions on behalf of the calling application. The security properties of an
+HCL integration are determined primarily by how the library is used, not by the library
+itself.
+
+This document describes the integration security model for Go developers embedding HCL
+in their own applications. It is not exhaustive. Integrators are responsible for assessing
+their own risk posture and applying controls appropriate to their deployment context.
+
+> **HCL configuration from untrusted sources must not be evaluated unless appropriate
+> security controls are in place.**
+
+## What HCL Guarantees
+
+The HCL library itself makes the following guarantees in isolation:
+
+- The expression evaluator does not make network calls, read environment variables, or
+  execute system commands.
+- HCL reads files from disk during the **parsing** step when loading configuration
+  source, but the evaluator itself performs no I/O.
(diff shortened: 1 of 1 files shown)


apparentlymart:
One thing that might be worth noting is that neither the parser nor evaluator currently have any stack overflow protection, and so multi-tenant applications parsing or evaluating untrusted HCL source code can potentially experience unrecoverable stack overflow errors, which could potentially be used for denial of service.

HCL (both in its original form that we now call "HCL 1", and in its current form "HCL 2") was written under the assumption that the parsing and evaluation would happen in a situation where the user causing a stack overflow would only be ruining their own experience and not affecting any other concurrent users.

That's definitely the case in software like Terraform and Packer where a process runs on behalf of only a single trusted operator. I believe the "server-like" HCL callers like Nomad achieve this by doing the HCL parsing client-side (e.g. in the CLI) and submitting only the resulting data structures to the server, and so again a stack overflow only crashes a process running on behalf of the single user.

---

For the parser I believe folks have already previously reported the possibility of stack overflow by trivial input such as long strings of parenetheses that'll just keep recursively entering the "parse expression" function.

I don't remember anyone directly reporting the evaluator equivalent, and I expect it _is_ harder to achieve a crash there without crashing the parser first, but there are probably ways to do it with creative uses of the more complex expression types like `for` expressions and the splat operator.

mildwonkey (approved):

mildwonkey:
@apparentlymart: you are, of course, correct - a stack overflow is what lead to this document's existence - but we're going to stick with being as vaguely generic as possible here. You can fill in the blanks as to our motivations 😂 (I'm not opposed to it - we can always add more detail later)

apparentlymart:
For what it's worth: the current text reads to me as saying that all of the risk is in how you populate _`hcl.EvalContext` in particular_: it seems specific but incomplete, rather than just being vague.

My intention was to say that passing untrusted source code to the parser functions is _also_ a risk, in addition to the risks this document is already describing. So one possible way to respond to that is to add "the input buffer or" before every reference to the evaluation context, making it still specific but more complete. Another way would be to stop referring to the evaluation context in particular and just refer to "untrusted input" _in general_, regardless of what part of the API it's being passed to, thereby making it more vague.

Of course I don't know what the actual goals were for this documentation, so if that feedback isn't useful then no worries! :man_shrugging:

PR #830, opened 2026-09-09
json: point mismatched delimiter diagnostics at offending token

cuishuang:
## Description

Fix JSON parser diagnostics for mismatched closing delimiters.

When parsing an object closed with `]`, or an array closed with `}`, the parser consumed the offending token and then used `p.Peek()` for the diagnostic subject. This caused the diagnostic range to point at the following token or the recovery position instead of the mismatched delimiter itself.

This change preserves the consumed token and uses its range as the diagnostic subject. Regression tests cover both object and array cases.

## Related Issue

Not applicable.

## How Has This Been Tested?

- `go test ./json`
- `go test ./...`
- `go vet ./...`
- `make fmtcheck`

### DIFF
--- a/json/parser.go
+++ b/json/parser.go
@@ -222,12 +222,12 @@ Token:
 		case tokenBrackC:
 			// Consume the bracket anyway, so that we don't return with the peeker
 			// at a strange place.
-			p.Read()
+			tok := p.Read()
 			return nil, diags.Append(&hcl.Diagnostic{
 				Severity: hcl.DiagError,
 				Summary:  "Mismatched braces",
 				Detail:   "A JSON object must be closed with a brace, not a bracket.",
-				Subject:  p.Peek().Range.Ptr(),
+				Subject:  &tok.Range,
 			})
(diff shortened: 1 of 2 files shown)


crw:
Hi, thanks for the submission. As this bypassed the normally required process as outlined in [CONTRIBUTING.md](https://github.com/hashicorp/terraform/blob/main/.github/CONTRIBUTING.md), particularly [Proposing a change](https://github.com/hashicorp/terraform/blob/main/.github/CONTRIBUTING.md#proposing-a-change) and potentially [AI Usage](https://github.com/hashicorp/terraform/blob/main/.github/CONTRIBUTING.md#ai-usage), I am going to close this pull request. I'd suggest first, in the issue, establishing if the maintainers are open to reviewing a change in this area and then following the guidelines in Contributing.md. Thanks again!

Note that I am referring to the Terraform contribution guidelines. We have not yet updated HCL with similar guidelines but should be doing so shortly. In this case, the same reasoning applies.

PR #834, opened 2026-09-14
github: Add consistency test for generated code

radeksimko:
## Description

This aims to ensure that generated code (as produced by `go generate ./...`) remains consistent and PRs don't change source without re-generating the code or cause other similar issues.

The workflow was originally motivated by discussion in https://github.com/hashicorp/hcl/pull/814

## Welcome to the 🐰 🕳️ 

Running a slightly modified version of this workflow in my personal fork revealed that Ragel (or at least some versions that I tested) has unfortunately non-deterministic output on all platforms but Linux.

 - https://github.com/radeksimko/hcl/pull/4
 - https://github.com/radeksimko/hcl/pull/5
 - https://github.com/radeksimko/hcl/pull/6

It will produce a few different offsets in `_hcltok_trans_keys` and `_hcltok_indicies` in `hclsyntax/scan_tokens.go` if executed repeatedly on the exact same macOS or Windows machine. Additionally, it produces some different LOC within a few `//line` directives depending on which OS it runs on.

It's unclear to me if the bug is triggered by any specific input that goes into the compiler in any specific Ragel file. I suspect it is, considering we have other Ragel generated code which does not suffer from this but I didn't have enough time to narrow it down to a minimal repro case yet.

## Proposal

Non-deterministic output from Ragel is most likely caused by upstream bug(s) which will need to be reported and (hopefully) fixed.

In the meantime I'm proposing that we use Linux runner in CI to produce and verify canonical form of the generated Ragel output and we can move on with https://github.com/hashicorp/hcl/pull/814 and regain some confidence about the generated code.

I'm also adding `.ruby-version` to pin the version of Ruby for that one Ruby script in https://github.com/hashicorp/hcl/blob/main/hclsyntax/unicode2ragel.rb - I think ideally in the long run this should be re-written in Go but that would be a separate stream of work. At least this way we can have more predictable/stable output for now.


.github/actions/setup-ragel/action.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    - name: Install Ragel (macOS)
+      shell: bash
+      if: ${{ runner.os == 'macOS' }}
+      run: brew install ragel
+
+    - name: Install Ragel (Windows)
+      if: ${{ runner.os == 'Windows' }}
+      shell: pwsh
+      run: winget install -e --accept-source-agreements --id PolarGoose.Ragel
```
radeksimko, line 23: This is effectively dead code for the time being but I'm leaving it here under the optimistic assumption that Ragel is fixed eventually and we can use this to continuously verify the output.

It seems simple enough to keep IMHO but if there are strong feelings about this I can also remove it 🤷🏻‍♂️

hclsyntax/scan_string_lit.go
```diff
@@ -146,11 +146,11 @@ func scanStringLit(data []byte, quoted bool) [][]byte {
 	    ret = append(ret, data[ts:te])
 	}*/
 
-//line scan_string_lit.go:156
+//line scan_string_lit.go:151
 	{
 	}
 
-//line scan_string_lit.go:160
+//line scan_string_lit.go:154
```
radeksimko, line 153: The LOC appears to be off here - both before and after but it's only informative for the debugger + it is likely another/same upstream bug we can do nothing about here.

hclsyntax/scan_string_lit.go
```diff
@@ -274,7 +274,7 @@ func scanStringLit(data []byte, quoted bool) [][]byte {
 					te = p
 					ret = append(ret, data[ts:te])
 
-//line scan_string_lit.go:280
+//line scan_string_lit.go:274
```
radeksimko, line 277: This is also off by a few lines, both before and after.

hclsyntax/scan_string_lit.go
radeksimko: Though maybe it is closer than before - I'm still not sure which exact line it is supposed to point to. 🤷🏻

SarahFrench: Same here - I wonder if adding copyright headers contributed at all? But even when I look at [the file before that change](https://github.com/hashicorp/hcl/blob/6c4344623b6ac528a57f9b80e4622acfab2fde40/hclsyntax/scan_string_lit.go) I'm unsure what I'm looking at.

radeksimko: AFAIK the copywrite tool ignores some generated files, which was also the source of the issue fixed in https://github.com/hashicorp/hcl/pull/832

radeksimko: Oh you mean _that_ changes may have contributed to the LOC changes here - possibly 🤔

.github/workflows/consistency-test.yml
```diff
@@ -0,0 +1,65 @@
...
+            git diff > go-generate.patch
+            exit 1
+          fi
+
+      - name: Upload patch
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          path: go-generate.patch
+          name: patches-${{ runner.os }}-${{ runner.arch }}
```
SarahFrench, line 65: Would the expected workflow be that someone not using Linux locally would obtain this patch after an inconsistency is flagged and apply it in a new commit?

hclsyntax/scan_string_lit.go
SarahFrench: Yeah exactly that

Either way, I agree it only impacts stuff like debuggers so we don't need to solve the mystery now. And arguably it's been a little bit off for a while now!

.github/workflows/consistency-test.yml
radeksimko: Yes, precisely - that is the expectation!

SarahFrench (approved):
This is a hairy looking problem, but this solution looks like the best we can do to ensure consistency while navigating a problem we don't have full power to resolve.

I don't know all the ins and outs of Ragel for codegen but I'm satisfied that:
- There is a clear workflow identifying inconsistencies and updating code to match 'what it should be'.
- The annotations with changing line references aren't impactful except in specific situations, and even then the risk is only a bit of confusing output from a debugger or code coverage tool.

PR #835, opened 2026-09-15
hclwrite: `NewExpression(Raw|Literal)` parses the expression and fallback to the original behavior

magodo:
## Description

With the addition of #810, the `hclwrite` is now able to parse expressions to get a _rich_ `Expression` that the users can further cast into specified expression types to further mutate them with the merge of #811.

As the process of mutating a parsed body (or even build up from scratch), the user shall be able to set expressions via `SetAttribute(Raw|Literal)`, whose return value (i.e. `Attribute`) shall keep the semantic of the expression being set, which means the `Attribute.Expr()` shall return the _rich_ `Expression` rather than the unstructured one.

~This PR introduces a `parseExpr` (similar to the `parse()`), which parses an expression via the extended `ParsaExpression()` in #810. In case it failed, i.e. the input raw tokens/literal value is not a valid expression, it falls back to the previous behavior to construct the unstructured tokens as an expression, to avoid potential breaking changes.~

**Updated** In order to avoid any potential breaking changes, this PR adds two new public functions:

- `ParseExpression`: Similar to `ParseConfig`, but for parsing expressions.
- `Body.SetAttribute`: This method sets a `*Expression` to an attribute, which can be either semantic rich (derived from `ParseExpression`) or not.

Note that with this direction, the #811 will also need to add a `SetItem()` accordingly.

I've also added UT for the `parseExpr` mainly to just copying the added test cases for `parse` in #810 to cover a variaty of expressions.

Note that this PR depends on #810.

## How Has This Been Tested?

```shell
hcl magodo/parse_expr
❯ go test ./...
ok  	github.com/hashicorp/hcl/v2	(cached)
?   	github.com/hashicorp/hcl/v2/cmd/hcldec	[no test files]
?   	github.com/hashicorp/hcl/v2/cmd/hclfmt	[no test files]
?   	github.com/hashicorp/hcl/v2/cmd/hclspecsuite	[no test files]
?   	github.com/hashicorp/hcl/v2/ext/customdecode	[no test files]
ok  	github.com/hashicorp/hcl/v2/ext/dynblock	0.009s
ok  	github.com/hashicorp/hcl/v2/ext/transform	0.003s
ok  	github.com/hashicorp/hcl/v2/ext/tryfunc	0.007s
ok  	github.com/hashicorp/hcl/v2/ext/typeexpr	0.010s
ok  	github.com/hashicorp/hcl/v2/ext/userfunc	0.010s
ok  	github.com/hashicorp/hcl/v2/gohcl	0.008s
ok  	github.com/hashicorp/hcl/v2/hcldec	0.011s
?   	github.com/hashicorp/hcl/v2/hcled	[no test files]
?   	github.com/hashicorp/hcl/v2/hclparse	[no test files]
ok  	github.com/hashicorp/hcl/v2/hclsimple	0.009s
ok  	github.com/hashicorp/hcl/v2/hclsyntax	0.031s
ok  	github.com/hashicorp/hcl/v2/hclsyntax/fuzz	0.005s
ok  	github.com/hashicorp/hcl/v2/hcltest	0.004s
ok  	github.com/hashicorp/hcl/v2/hclwrite	0.083s
ok  	github.com/hashicorp/hcl/v2/hclwrite/fuzz	0.005s
ok  	github.com/hashicorp/hcl/v2/integrationtest	0.005s
ok  	github.com/hashicorp/hcl/v2/json	0.007s
ok  	github.com/hashicorp/hcl/v2/json/fuzz	0.004s
ok  	github.com/hashicorp/hcl/v2/specsuite	0.516s
```

### DIFF
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2014, 2025
+// Copyright IBM Corp. 2014, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package hclwrite
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -187,6 +187,18 @@ Traversals:
 	}
 }
 
+func (e *Expression) AsObjectConsExpr() *ObjectConsExpr {
+	var found *ObjectConsExpr
+	e.walkChildNodes(func(n *node) {
+		if o, ok := n.content.(*ObjectConsExpr); ok {
+			found = o
+			return
+		}
+	})
+
+	return found
+}
+
 // Traversal represents a sequence of variable, attribute, and/or index
 // operations.
 type Traversal struct {
--- a/hclwrite/ast_expression.go
+++ b/hclwrite/ast_expression.go
@@ -15,6 +15,7 @@ type Expression struct {
 	inTree
 
 	absTraversals nodeSet
+	wrapped       *node
 }
 
 func newExpression() *Expression {
@@ -22,6 +23,15 @@ func newExpression() *Expression {
 		inTree:        newInTree(),
 		absTraversals: newNodeSet(),
 	}
+}
+
+func (e *Expression) wrap(c nodeContent) {
+	if e.wrapped != nil {
+		panic("this is a problem")
+	}
+
+	e.wrapped = newNode(c)
+	e.children.AppendNode(e.wrapped)
 }
 
 // NewExpressionRaw constructs an expression containing the given raw tokens.
@@ -188,15 +198,35 @@ Traversals:
 }
 
 func (e *Expression) AsObjectConsExpr() *ObjectConsExpr {
-	var found *ObjectConsExpr
-	e.walkChildNodes(func(n *node) {
-		if o, ok := n.content.(*ObjectConsExpr); ok {
-			found = o
-			return
-		}
-	})
+	if e.wrapped == nil {
+		return nil
+	}
 
-	return found
+	if object, ok := e.wrapped.content.(*ObjectConsExpr); ok {
+		return object
+	}
+
+	return nil
+}
+
+func (e *Expression) asQuoted() *quoted {
+	if e.wrapped == nil {
+		return nil
+	}
+
+	if quoted, ok := e.wrapped.content.(*quoted); ok {
+		return quoted
+	}
+
+	return nil
+}
+
+func (e *Expression) AsQuotedLiteral() Tokens {
+	if quoted := e.asQuoted(); quoted == nil {
+		return Tokens{}
+	} else {
+		return quoted.tokens
+	}
 }
 
 // Traversal represents a sequence of variable, attribute, and/or index
--- a/hclwrite/ast_object_cons_expr.go
+++ b/hclwrite/ast_object_cons_expr.go
@@ -18,8 +18,16 @@ func (o *ObjectConsExpr) ValueFor(key string) *ObjectConsValue {
 	o.walkChildNodes(func(n *node) {
 		if item, ok := n.content.(*ObjectConsItem); ok {
 			k := item.key.content.(*ObjectConsKey)
-			name := k.name.content.(*identifier)
-			if k.literal && string(name.token.Bytes) == key {
+			name := k.name.content.(*Expression)
+
+			// A temporary workaround for parsing expressions as names at ...
+			// parse-time
+			maybeKey := ""
+			for _, token := range name.BuildTokens(nil) {
+				maybeKey += string(token.Bytes) // no spaces before
+			}
+
+			if k.literal && (maybeKey == key || maybeKey == `"`+key+`"`) {
 				found = item.value.content.(*ObjectConsValue)
 				return
 			}
--- a/hclwrite/ast_object_cons_expr.go
+++ b/hclwrite/ast_object_cons_expr.go
@@ -1,0 +1,76 @@
+package hclwrite
+
+type ObjectConsExpr struct {
+	inTree
+}
+
+func newObjectConsExpr() *ObjectConsExpr {
+	return &ObjectConsExpr{
+		inTree: newInTree(),
+	}
+}
+
+func (o *ObjectConsExpr) ValueFor(key string) *ObjectConsValue {
+	var found *ObjectConsValue
+	o.walkChildNodes(func(n *node) {
+		if item, ok := n.content.(*ObjectConsItem); ok {
+			k := item.key.content.(*ObjectConsKey)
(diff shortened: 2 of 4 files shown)


apparentlymart:
Having `NewExpressionRaw` try to treat the given tokens as representing an expression seems like reasonable behavior to me, since it means that if successful the resulting `*Expression` can have all of the other analyses performed on it. There is some risk of breaking change here since converting a `Tokens` to `[]byte` loses the information about which token types were used in the input, but it'd be breaking only if the caller was using the wrong token type for a specific byte sequence and so hopefully the risk here is low.

I don't think changing `NewExpressionLiteral` in this way makes sense though, because the purpose of that function is to produce an expression that represents the given value _literally_ so that evaluating it would produce an equivalent `cty.Value`. Existing callers are expecting to be able to construct expressions representing strings that might contain HCL-expression-like syntax without that being misinterpreted as actually being an HCL expression. This is therefore a breaking change and removes important functionality.

Instead I'd suggest adding a new exported function:

```go
// ParseExpression attempts to parse the given buffer as a HCL expression,
// returning the resulting syntax element if successful.
func ParseExpression(src []byte) (*Expression, hcl.Diagnostics)
```

I'm intending for this to match with [`hclsyntax.ParseExpression`](https://pkg.go.dev/github.com/hashicorp/hcl/v2/hclsyntax#ParseExpression) in a similar way to how `hclwrite.ParseConfig` matches with `hclsyntax.ParseConfig`. A future PR could also add `hclwrite.ParseTemplate` to match with [`hclsyntax.ParseTemplate`](https://pkg.go.dev/github.com/hashicorp/hcl/v2/hclsyntax#ParseTemplate) and then all of the main parsing functions would have `hclwrite` counterparts.

magodo:
@apparentlymart The reason I chagned the `NewExpressionLiteral` the same way as I did to `NewExpressionRaw` is because it is used in `SetAttributeValue` (and the `SetItemValue` in #811). I didn't know there is a use of:

> Existing callers are expecting to be able to construct expressions representing strings that might contain HCL-expression-like syntax without that being misinterpreted as actually being an HCL expression

To have a new public function `ParseExpression` makes sense. While I still want to have a way to cover this scenario: A user constructs a `hclwrite.File` gradually, using functions like `AppendBlock()`, `SetAttributeXXX()`, and the newly added `SetItemXXX()` if the added attribute is a ObjectConsExpr (as well as for the TupleConsExpr in the future). Without this change, all the newly added attribute is just a bunch of tokens, which means the users can't mutate the inner objects.

Making `SetAttributeRaw()` to keep the semantic, but leave the `SetAttributeValue()` to not keep it seems inconsistent. WDYT?

apparentlymart:
I think the missing primitive here is:

```go
func (b *Body) SetAttribute(name string, expr *Expression) *Attribute
```

I left this out of the original implementation because in practice there weren't any useful ways to create a standalone `*Expression`, and the only caller we were worrying about at the time (Terraform) only needed to be able to set literal values and traversals.

But now that you're adding the missing `ParseExpression` there will be a general way to produce arbitrary `*Expression` values, and so `Body.SetAttribute` would now actually be useful. `Body.SetAttributeLiteral` and `Body.SetAttributeTraversal` are helpful shortcuts for the two most common cases, but callers that want to do more complicated things can do it in two steps by first parsing the expression and then assigning it to the attribute.

```go
expr, diags := hclwrite.ParseExpression(src)
// (handle diagnostics)
body.SetAttribute("example", expr)
```

---

`SetAttributeValue` should continue supporting only literals for the same reason why `NewExpressionLiteral` should continue only producing literals: callers rely on it to be able to set arbitrary values for attributes with them being automatically escaped if they contain HCL template syntax.

[OpenTofu's implementation of `encode_tfvars`](https://opentofu.org/docs/v1.12/language/providers/builtin/#encode_tfvars) is an example caller that relies on this, to ensure that any string passed to it is properly escaped in the resulting `.tfvars`-style string. Terraform's equivalent function presumably does something similar.

magodo:
I spent sometime to understand what does the following really mean:

> Existing callers are expecting to be able to construct expressions representing strings that might contain HCL-expression-like syntax without that being misinterpreted as actually being an HCL expression

Per my understanding, a `cty.Value` is constructed by a parsed `hclsyntax.Expression`, calling its `Value()` method. Conditionally, for cty types like List/Set, since the `NewExpressionLiteral` doesn't treat them the same, one will need to additionally call `convert.Convert(val, typ)` to convert it back to the expected type, which is also called out in the comment of `NewExpressionLiteral`:

```
// ...
// HCL native syntax does not directly represent lists, maps, and sets, and
// instead relies on the automatic conversions to those collection types from
// either list or tuple constructor syntax. Therefore converting collection
// values to source code and re-reading them will lose type information, and
// the reader must provide a suitable type at decode time to recover the
// original value.
func NewExpressionLiteral(val cty.Value) *Expression {
```

The `hclsyntax.Expression` comes from `hclsyntax.ParseExpression`, which takes a `[]byte`, which comes from `hclwrite.Expression.BuildTokens(nil).Bytes()`. As it consumes the `[]byte`, the token types doesn't really matter here ()?

Also, for HCL-like cty.String, even if the tokens are converted to bytes for the inner parsing, isn't it still ending up to be a string, instead of some other HCL expression?

magodo:
@apparentlymart I agree that introducing `ParseExpression`, `SetAttribute` and the `SetItem` (in #811) shall solve the use case I've described above.

I still struggle to understand how does my current implementaion break the `NewExpressionLiteral` when the input string is a HCL-like cty.String? In my version, I still call the `TokensForValue()`, which escapes bytes as necessary.

apparentlymart:
Okay, reading back what you wrote I think I misunderstood what you changed in the literal-related functions. Sorry about that.

I initially thought that what was happening was that the actual value of a `cty.String` would be parsed as an expression. But I think I understand now that your intention is to generate the tokens that represent a string literal -- the open quote, the literal string value, and then the close quote -- and parse the byte buffer representation of those three tokens into a proper syntax tree instead of just the three flat unstructured tokens.

And that does seem like a reasonable thing to do. I'm sorry for giving confusing advice due to my misunderstanding.

magodo:
@apparentlymart Thank you for the clarification! I actually reverted my previous changes and pushed a few commits per your suggestion (with the PR description updated as well). I'll keep the commit as is for now and leave it to the maintainer's to decide.

bbasata:
Looks good to me, as author of #810 and #811. I will defer to the maintainers for review. Acknowledging @magodo's notes:

> Note that with this direction, the https://github.com/hashicorp/hcl/pull/811 will also need to add a `SetItem()` accordingly.

> Note that this PR depends on https://github.com/hashicorp/hcl/pull/810.

PR #837, opened 2026-09-18
hclsyntax: include leading comments in top-level body SrcRange

### LINKED ISSUE #551: Incorrect `srcRange` for files starting with a multiline comment

adityabagla7:
## Summary
- Fixes #551: top-level `Body.SrcRange` now starts at the first token in the file, including leading multiline comments that the peeker filters out.
- Nested bodies are unchanged (they still start from the already-consumed opener via `PrevRange`).
- Adds a `TestParseConfig` case covering a leading `/* ... */` comment.

## Test plan
- [x] `go test ./hclsyntax/`
- [x] Reproduced the issue fixture: range is now `main.tf:1,1-8,1` (was `main.tf:3,34-8,1`)

Made with [Cursor](https://cursor.com)


crw:
Thanks for your interest. It is unlikely this will be reviewed at this time, I would suggest proposing your change on the issue and see if there is maintainer interest in reviewing a contributed solution. Thanks!

PR #838, opened 2026-09-20
fix: omit nil Subject prefix in Diagnostic.Error()

### LINKED ISSUE #399: Error formatting can result in "<nil>" in output
The conversion of a Diagnostic to an error string assumes a meaningful subject: https://github.com/hashicorp/hcl/blob/90676d47a0ce038f02596e2aa5894bf0c57fad56/diagnostic.go#L77

In practice I'm seeing errors such as `<nil>: Configuration file not found; The configuration file foo.hcl does not exist.` which are a bit ugly. Should the inclusion of `Subject` be conditional, or would that cause issues by breaking the current X:Y structure?

vulragrag-star:
## Description

`Diagnostic.Error()` always formatted `Subject` with `%s`. When a diagnostic has no source range (for example `hclsimple.DecodeFile` / `hclparse.ParseHCLFile` on a missing path), Go prints a literal `<nil>`:

```
<nil>: Configuration file not found; The configuration file foo.hcl does not exist.
```

Skip the subject prefix when `Subject` is nil, matching how `diagnostic_text.go` already treats a missing subject.

## Related Issue

Fixes #399

## How Has This Been Tested?

- Added `TestDiagnosticError` covering nil-subject and with-subject cases
- `go test -count=1 ./...` PASS


radeksimko:
I believe that the example mentioned (file not found) is a FS/IO error which should be treated as such and not wrapped around a HCL diagnostic because it has nothing to do with HCL or the parser in it.

I'm sure there are some valid use cases for a source-less diagnostics but this isn't one. We should first identify these use cases before accounting for them, and also reflect them in a test case.

Let's keep the conversation going in https://github.com/hashicorp/hcl/issues/399 before proposing a solution.

